PR #1, opened 2016-08-26
make tev source 49, change filenames of the others

vorugantia:
@cdeil - I made one tev source. If this looks good, I'll just go down the lists in the seed folder and see how many I can get through today.


input/sources/tev-000049.yaml
```diff
@@ -0,0 +1,10 @@
+source_id: 49
```
cdeil: This is a change. I used a `source_id` string, you're using an int.
I actually would prefer to keep this change.
Can you change the other files also?

input/sources/README.md
```diff
@@ -12,7 +12,7 @@ from the TeVCat and TeGeV cat webpages.
 
 Schema (examples from `tev-000234.yaml`):
 
-* `source_id` - Source ID in this catalog (str, e.g. `tev-000234`)
+* `source_id` - Source ID in this catalog (str, e.g. `tev-000234`). For now use tgevcat ID.
```
cdeil: Change this description to say it's an int.
And change the example ID ... if we change to tgevcat it's no longer 234.

input/sources/tev-000049.yaml
```diff
@@ -0,0 +1,10 @@
...
+
+tevcat_id: 75
+tevcat2_id: cIB7Uf
+tevcat_name: TeV J1104+382
+
+tgevcat_id: 49
+tgevcat_name: TeV J1104+3811
+
+papers:
```
cdeil: For the papers, I think it would be best to keep the `- none` from the example file.
It'll make it easier to add to the list of papers later (every source will have at least one paper)

cdeil:
Thanks!

I left some inline comments.

This mostly looks good.

My suggestion would be that you don't to do too many files today (say ~ 10), in case we notice we have to change some formatting.

If you want to practice some Python coding, you could try to implement the code that gathers the info from these files into a ECSV table, and then makes a HTML version with actual links to TeVCat 1 and 2. I can help with that tonight.

vorugantia:
Alright, I'll work on ~10 more and make another commit.

vorugantia:
Added 10 more sources. I'm going to get some lunch and then I can start with the Python coding,

cdeil:
Thanks!

PR #2, opened 2016-08-27
add the rest of the source yaml files

vorugantia:
In this PR, I've created the other yaml files for the sources. Some issues:

* Any tevcat id beyond 257 did not have a tevcat2 id (the URL doesn't redirect). For these sources, I left the tevcat2_id field as `null`.

* [tevcat id 129](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L64) is unpaired with any tgevcat source, but is SIMILAR to [tevcat id 74](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L16) (which is paired).

* tevcat id [163](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L89) and [164](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L90) are unpaired, because tgevcat "combines" these two with [tgevcat id 111](https://github.com/gammapy/gamma-cat/blob/master/seed/tgevcat/tgevcat.ecsv#L173).

* [tevcat id 261](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L146) is unpaired. One of its alternative names does appear on [tgevcat id 145](https://github.com/gammapy/gamma-cat/blob/master/seed/tgevcat/tgevcat.ecsv#L207), but this is paired with ANOTHER tevcat source ([id 149](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L79)).

* Same issue as above for both tevcat id [278](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L148) and [279](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L149) - they match up with [tgevcat id 130](https://github.com/gammapy/gamma-cat/blob/master/seed/tgevcat/tgevcat.ecsv#L192), but that's already paired with [tevcat id 143](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L74).

* Same issue again with [tevcat id 253](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L175) - it matches with [tgevcat id 37](https://github.com/gammapy/gamma-cat/blob/master/seed/tgevcat/tgevcat.ecsv#L99), but that's already paired with [tevcat id 86](https://github.com/gammapy/gamma-cat/blob/master/seed/tevcat/tevcat.ecsv#L26).

* These are the other tevcat sources that simply do not have a tgevcat counterpart at all: 88, 174, 263-274, 277, 280.

Sorry if that's a bit confusing.

The reason I couldn't fill in most of these missing fields as "null" is because if there isn't a tgevcat source to match, I don't know what to call the file (our id depends on tgevcat's source id).

Other than that, all of the files in this PR should be just fine. I did check them over throughout to make sure I wasn't making any mistakes.

### DIFF
--- a/input/sources/tev-000001.yaml
+++ b/input/sources/tev-000001.yaml
@@ -1,0 +1,11 @@
+source_id: 1
+
+tevcat_id: 227
+tevcat2_id: rC5JCj
+tevcat_name: TeV J0006+729
+
+tgevcat_id: 1
+tgevcat_name: TeV J0006+7259
+
+papers:
+- null
--- a/input/sources/tev-000002.yaml
+++ b/input/sources/tev-000002.yaml
@@ -1,0 +1,11 @@
+source_id: 2
+
+tevcat_id: 213
+tevcat2_id: 70hRXZ
+tevcat_name: TeV J0013-188
+
+tgevcat_id: 2
+tgevcat_name: TeV J0013-1853
+
+papers:
+- null
--- a/input/sources/tev-000003.yaml
+++ b/input/sources/tev-000003.yaml
@@ -1,0 +1,11 @@
+source_id: 3
+
+tevcat_id: 209
+tevcat2_id: xCyAjF
+tevcat_name: TeV J0025+641
+
+tgevcat_id: 3
+tgevcat_name: TeV J0025+6410
+
+papers:
+- null
--- a/input/sources/tev-000004.yaml
+++ b/input/sources/tev-000004.yaml
@@ -1,0 +1,11 @@
+source_id: 4
+
+tevcat_id: 239
+tevcat2_id: PKJZqs
+tevcat_name: TeV J0033-193
+
+tgevcat_id: 4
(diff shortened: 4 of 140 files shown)


cdeil:
This is awesome. Thank you!

I'm merging this now.

There's a few YAML file formatting issues pointed out on travis-ci that I'll fix in master:
https://travis-ci.org/gammapy/gamma-cat/builds/155533192

Concerning your comment about differences between the TeV catalogs: thanks for being so thorough. We'll have to look at and decide on each one by itself. But I don't want to do this now, we'll develop our criteria for what's a TeV source over time. Still, I'll make a new Github issue pointing to this one just so that we don't forget to look at these cases.

PR #13, opened 2016-09-05
Add more sources

vorugantia:
I added some more sources, I'm up to 130 right now. I might have some time later today to do more. (I also added Vincent's new paper to source_id 133, W 39B).


cdeil:
Thank you!

Merging this now.

PR #33, opened 2016-10-31
Add SED class and check input SEDs

cdeil:
This PR:
- Adds an SED and SEDList class to gather and read the input SEDs.
- Fixes misc issues with the YAML header and data part of 12 input SEDs
- Adds input SED validation to `make.py check`, i.e. from now on SEDs should be OK.


cdeil:
@adonath - There were several formatting issues with input SEDs.
They are fixed in this PR and from now on `./make.py check` will check the format of all SEDs, so from now on things should be better.

PR #31, opened 2016-11-02
Migrate data format of SEDs

cdeil:
This PR migrates the SEDs to the new data formats:

https://github.com/gammapy/gamma-cat/blob/master/input/migrations/migrate_seds.py
http://gamma-astro-data-formats.readthedocs.io/en/latest/results/flux_points/index.html

I could have done it by hand, but I wanted to see if it's possible to round-trip input files via Python scripts. Knowing how to do that will be useful in the future, if we ever want to change the format of input files after having already collected a lot of data.

I've put some notes about schema migrations here:
https://github.com/gammapy/gamma-cat/blob/master/input/migrations/README.md


cdeil:
@adonath - Could you please review this PR?
Did I fill the `dnde_errn` and `dnde_errp` fields correctly?
Are the auto-changes to the ECSV files acceptable?

adonath:
@cdeil It seems like the first row of the old flux points is consistently missing in the new files (or overwritten by the header row...). Can you check what happened there?

cdeil:
@adonath - Thanks for noticing the issue with the header / first data line!

I'm abandoning this PR, closing it now.

I've cherry-picked some things into master, and will continue applying manual fixes there (namely adding the CSV header line to the SEDs manually and adding a check that it's correct.

Not sure yet if I want to use this migration script at all, or apply the renames manually.

cdeil:
@adonath - FYI: I'm now going through and fix all SED files by hand. Should be done in an hour.

cdeil:
I added a test for the CSV header line:
https://github.com/gammapy/gamma-cat/blob/master/gammacat/sed.py#L32
and finished the reformatting.

I'll now add scripts to produce SEDs (with uniform fields, units, description) in the output folder.

PR #34, opened 2016-11-23
Spectral points for 2006ApJ...636..777A HESS J1745-303

cdeil:
@adonath - I noticed that there's one source from the 2006 HESS survey paper (HESS J1745-303, source_id = 108) where the spectral points are missing:
https://github.com/gammapy/gamma-cat/tree/master/input/papers/2006/2006ApJ...636..777A

Is there a reason why they are missing?
Or are they available and should still be added?


cdeil:
I've checked, the lo / hi SED errors were switched for all ECSV files for this paper.
Fixed in 7cff1ed .

I didn't check SEDs for other papers for this mistake.
Would be good of course, but IMO the chances that all is good are pretty high.
So instead I'll continue adding the missing SEDs here and for other papers now.

cdeil:
This is complete. I'm merging this now.

I've checked that for all sources at  https://www.mpi-hd.mpg.de/hfm/HESS/pages/publications/auxiliary/ApJ_636.html  there's a YAML and ECSV SED file now in https://github.com/gammapy/gamma-cat/tree/master/input/papers/2006/2006ApJ...636..777A .

PR #25, opened 2016-11-23
Add 2016Natur.531..476H, HESS Galactic center

cdeil:
Add data from:

http://adsabs.harvard.edu/abs/2016Natur.531..476H
http://www.nature.com/nature/journal/v531/n7595/fig_tab/nature17147_F3.html

Spectral points (obtained via private communication from Aion Viana on Oct 19)
https://gist.github.com/cdeil/a4fd2afe33b232e14e709a49f254ad08


cdeil:
I added the data here. Merging this now.

cdeil:
I fixed the SED upper limit points for this dataset just now in f284ff7 .

PR #35, opened 2016-11-24
Add HESS, W49B, 2016arXiv160900600H

cdeil:
We should add the data for HESS, W49B, 2016arXiv160900600H
http://adsabs.harvard.edu/abs/2016arXiv160900600H

Spectral points from Francois Brun via email Nov 4, 2016:
https://gist.github.com/cdeil/48dcae6fa8bcf832c4328520a991539f

> The csv file has 6 columns :
Energy, Energy_errorlow, Energy_errorhigh, differential flux (df), df_errorlow, df_errorhigh

> The differential flux is given in cm-2.s-1.TeV-1 and the errors on the flux are statistical only.
The last point has an error of 0, this is an UL at 95% CL.

> P.S. : In the figures of the paper, systematic error of 20% of the flux value has been added in quadrature to the statistical error of each point.

### DIFF
--- a/input/papers/2016/2016arXiv160900600H/tev-000133.ecsv
+++ b/input/papers/2016/2016arXiv160900600H/tev-000133.ecsv
@@ -1,0 +1,23 @@
+# %ECSV 0.9
+# ---
+# datatype:
+# - {name: e_ref, datatype: float32}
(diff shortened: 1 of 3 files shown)


cdeil:
Done. See attached commits. Merging this now.

PR #37, opened 2016-11-24
Add HESS Vela Junior 2016arXiv161101863H

cdeil:
We should add the new Vela Junior paper info:
http://adsabs.harvard.edu/abs/2016arXiv161101863H

Spectral points and parameters from Manuel Paz Arribas via email on Oct 21, 2016 are here:
https://gist.github.com/cdeil/911a5f2d7f41fbe224d4783a5eb47a55


cdeil:
Done. Merging this now.

PR #48, opened 2016-12-08
HESS Vela X 2012 publication SED incorrect 

cdeil:
@adonath and I noticed that the SED for Vela X here is incorrect (or at least the lower limit):

https://github.com/gammapy/gamma-cat/blob/1815c4f14d3e369356f0e55d7885cd4792b6b051/input/papers/2012/2012A%2526A...548A..38A/tev-000037.ecsv

Paper: https://ui.adsabs.harvard.edu/#abs/2012A%26A...548A..38A

There's no SED points given on the HESS Aux page:
https://www.mpi-hd.mpg.de/hfm/HESS/pages/publications/auxiliary/AA548_A38.html

I'll have a look now and try to fix this...


cdeil:
My assumptions about the columns in the CSV file were wrong.
I figured out what's wrong and fixed it and left a comment in the header showing that it's correct now.

The commit is attached here so that there's an easy record.
Merging this now.

PR #71, opened 2017-02-02
Added Data to source 37

pdeiml:
See commit description

PR #72, opened 2017-02-02
2015 mnras.446.1163 h

pdeiml:
I think in the 2015 paper, there is no sigma/sigma2/pa for the gauss.


cdeil:
About `input/data/2006/2006A%26A...448L..43A/tev-000037.yaml` we talked yesterday, it needs the norm parameter for the spectral model, and the extension info for the Gauss morphology.

Wait ... I think Vela X is in a separate PR #71 and you accidentally also put the commit in this branch?
(e.g. by accidentally doing `git checkout -b 2015MNRAS.446.1163H` while in the `add_data` branch, where you should have done it from `master`?)

To get rid of it you can run `git rebase -i master` and delete that line with the commit you want to remove from this branch.

cdeil:
Concerning morph for 2015MNRAS.446.1163H, you're right, no `sigma` for extension is given.

They say that extension is not significance and give an upper limit on sigma at 0.074 deg at a 99 per cent confidence level (C.L.).

I don't think that I've added the possibility for upper limits yet to the schema. For now, could you please put this?
```
# TODO: add upper limit `sigma < 0.074 deg` at CL=0.99
sigma: 0
```
Then later we'll decide on a solution how to handle upper limits and via the `TODO` marker will be able to find this and clean it up.

We'll have to do this from time to time .... defer things to future pull request, so that data entry pull requests don't get stuck.

pdeiml:
Fehler überprüft und richtig eingegeben.
Ich habe einfach aus dem aktuellen masterbranch die VelaX-yaml-Datei in diesen Branch kopiert und hinzugefügt. Also Vela X ist jetzt auf dem gleichen Stand wie im Masterbranch.

Ready to merge.

cdeil:
Thanks!

PR #73, opened 2017-02-02
added data to 2015A%26A...577A.131H/tev-000045.yaml

pdeiml:
I didn't find err_sys for the position

### DIFF
--- a/input/data/2015/2015A%26A...577A.131H/tev-000045.yaml
+++ b/input/data/2015/2015A%26A...577A.131H/tev-000045.yaml
@@ -1,4 +1,20 @@
 source_id: 45
 reference_id: 2015A&A...577A.131H
 
-# TODO: add data
+data:
+  livetime: 63.3
+  significance: 9.3
+
+pos:
+  ra: {val: 10h18m58s, err: 0h0m5s, }
+  dec: {val: -58d56m43s, err: 0d0m30s, }
+
+morph:
+  type: point
+
+spec:
+  type: pl
+
+  norm: {val: 0.29, err: 0.04, err_sys: 0.058}
+  index: {val: 2.20, err: 0.14, err_sys: 0.2}
+  ref: 1


cdeil:
With the RA and DEC errors you have to be careful.
We process those values by parsing the angles via `astropy.coordinates.Angle`, and this is what it does:
```
In [2]: Angle('5s')
Out[2]: <Angle 5.0 arcsec>

In [3]: Angle('0h0m5s')
Out[3]: <Angle 0.001388888888888889 hourangle>

In [4]: Angle('0h0m5s').to('arcsec')
Out[4]: <Angle 74.99999999999999 arcsec>
```
So there's a difference between "s" meaning arcsec, and "s" being the seconds part of an hourangle. (24 hours, but 360 deg).

I think this needs to be fixed here, no?

Concerning systematic errors: they are sometimes given, and often not given.
When not given, we just put the statistical error in the file, just like you did here.

input/data/2015/2015A%26A...577A.131H/tev-000045.yaml
```diff
@@ -6,8 +6,8 @@ data:
   significance: 9.3
 
 pos:
-  ra: {val: 10h18m58s, err: 5s, }
-  dec: {val: -58d56m43s, err: 30s, }
+  ra: {val: 10h18m59.3s, err: 2.4s, }
+  dec: {val: -58d56m10s, err: 36s, }
```
pdeiml: Ich hatte eine Zahlendreher drin, aber ich habe es jetzt analog zum Paper 

https://arxiv.org/pdf/0705.1605v1.pdf

das zur yaml-Datei 

https://github.com/gammapy/gamma-cat/blob/master/input/data/2007/2007A%2526A…472..489A/tev-000115.yaml

gehoert eingetragen.

cdeil:
I think the position you have is from section 1 "intro" from a previous paper.
The measured position in this paper is given in section 2 and is different.

pdeiml:
Ok, ready

pdeiml:
Rechecked the errors on ra and dec, looks fine now.
Ready to merge!

cdeil:
Thanks!

PR #75, opened 2017-02-03
added data to 2014MNRAS.439.2828A/tev-000088.yaml

pdeiml:
found no excess value.

### DIFF
--- a/input/data/2014/2014MNRAS.439.2828A/tev-000088.yaml
+++ b/input/data/2014/2014MNRAS.439.2828A/tev-000088.yaml
@@ -1,4 +1,23 @@
 source_id: 88
 reference_id: 2014MNRAS.439.2828A
 
-# TODO: add data
+data:
+  livetime: 63.4
+
+pos:
+  ra: {val: 16h40m41.0s, err: 0h0m1.0s, err_sys: 0h0m1.3s}
+  dec: {val: -46d32m31s, err: 0d0m14s, err_sys: 0d0m20s}
+
+morph:
+  type: gauss
+  sigma: {val: 4.3m, err: 0.2m}
+
+spec: 
+  type: ecpl
+
+  norm: {val: 3.3, err: 0.1, err_sys: 0.6}
+  index: {val: 2.11, err: 0.09, err_sys: 0.10}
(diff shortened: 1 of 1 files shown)


input/data/2014/2014MNRAS.439.2828A/tev-000088.yaml
```diff
@@ -1,4 +1,23 @@
...
+  livetime: 63.4
+
+pos:
+  ra: {val: 16h40m41.0s, err: 1.0s, err_sys: 1.3s}
+  dec: {val: -46d32m31s, err: 14s, err_sys: 20s}
+
+morph:
+  type: gauss
+  sigma: {val: 4.3m, err: 0.2m}
+  sigma2: {val: 4.3m, err: 0.2m}
```
cdeil: Remove `sigma2` that duplicates the value of `sigma`.
The idea is that we only add `sigma2` for elongated sources, and here a symmetric Gaussian was fitted.
We could make our schema more explicit in the future, putting `gauss_elongated` or something, but for now this is how other sources were done.

input/data/2014/2014MNRAS.439.2828A/tev-000088.yaml
```diff
@@ -1,4 +1,23 @@
...
+
+spec: 
+  type: ecpl
+
+  norm: {val: 3.3, err: 0.1, err_sys: 0.6}
+  index: {val: 2.11, err: 0.09, err_sys: 0.10}
+  # TODO: Error on the cut-off energy is assymmetric: -1.2, +2.0
+  ecut: {val: 6.0, err: 2.0}
+  ref: 1
```
cdeil: Add newline at end of file (Github points this out in red in the diff, and doing it this way means that if others add lines in the future, the diff will be cleaner, i.e. not touch your last line here and potentially lead to fewer merge conflicts.

It's a super-minor point, but since there's in any case one thing to fix (see comment above), please do this change as well.

pdeiml:
check :-)

pdeiml:
Rechecked the errors.
Ready to merge!

cdeil:
Thanks!

PR #76, opened 2017-02-03
added data to 2014A%26A...562A..40H/tev-000117.yaml

pdeiml:
Im Absatz "Results" bin ich mir nicht sicher, ob da Infos über sigma/sigma2 drinstehen.

### DIFF
--- a/input/data/2014/2014A%26A...562A..40H/tev-000117.yaml
+++ b/input/data/2014/2014A%26A...562A..40H/tev-000117.yaml
@@ -1,4 +1,23 @@
 source_id: 117
 reference_id: 2014A&A...562A..40H
 
-# TODO: add data
+data:
+  livetime: 120
+  significance: 8.2
+  excess: 117
+
+pos:
+  ra: {val: 18h18m4.8s, err: 3.1s, err_sys: }
+  dec: {val: -15d28m1s, err: 43s, err_sys: }
+
+morph:
+  type: gauss
+
+spec:
+  type: pl
+
+  norm: {val: 0.09, err: 0.02, err_sys: 0.02}
+  index: {val: 2.3, err: 0.3, err_sys: 0.2}
+
+  erange: {min: 0.42, max: 12.0}
+
--- a/input/data/2014/2014A%26A...562A..40H/tev-000117.yaml
+++ b/input/data/2014/2014A%26A...562A..40H/tev-000117.yaml
@@ -7,17 +7,19 @@ data:
   excess: 117
 
 pos:
-  ra: {val: 18h18m4.8s, err: 3.1s, err_sys: }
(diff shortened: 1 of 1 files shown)


input/data/2014/2014A%26A...562A..40H/tev-000117.yaml
```diff
@@ -1,4 +1,23 @@
...
 reference_id: 2014A&A...562A..40H
 
-# TODO: add data
+data:
+  livetime: 120
+  significance: 8.2
+  excess: 117
+
+pos:
+  ra: {val: 18h18m4.8s, err: 3.1s, err_sys: }
```
cdeil: Section 2.3 states
> α = 18 18 4.8 ± 3.1stat

That means that the error on RA is 3 seconds (relative to an angle of 1 hour), not 3 arcseconds (which is seconds relative to an angle of 1 deg)
The correct way to put it is as string "0h0m0s" so that astropy Angle parses it correctly.

input/data/2014/2014A%26A...562A..40H/tev-000117.yaml
```diff
@@ -1,4 +1,23 @@
...
+  dec: {val: -15d28m1s, err: 43s, err_sys: }
```
cdeil: The paper states 
> In addi- tion to the statistical uncertainty of this fit, there is also a system- atic uncertainty due to the pointing precision of the telescope ar- ray of about 20′′

so you could add that info for err_sys = 20 arcsec for both axes

input/data/2014/2014A%26A...562A..40H/tev-000117.yaml
```diff
@@ -1,4 +1,23 @@
...
+
+morph:
+  type: gauss
```
cdeil: > The 99% confidence upper limit on the Gaussian width of the source is 0.072◦. No significant ellipticity of the source was found.

Please put this upper limit for `sigma` like in the other case where we had an upper limit on extension earlier today.

input/data/2014/2014A%26A...562A..40H/tev-000117.yaml
```diff
@@ -1,4 +1,23 @@
...
+
+spec:
+  type: pl
+
+  norm: {val: 0.09, err: 0.02, err_sys: 0.02}
+  index: {val: 2.3, err: 0.3, err_sys: 0.2}
+
```
cdeil: A decorrelation energy of `e_ref = 1.9 TeV` is mentioned in the paper.
You have to add this, or the spectrum will be incorrect (i.e. `e_ref =  1 TeV` would be assumed)

pdeiml:
Comments addressed. Ready to merge?

cdeil:
Looks good to me. Merging this now. Thanks!

PR #79, opened 2017-02-03
added data to 2009A%26A...499..723A/tev-000132.yaml

pdeiml:
Vergleich bitte mal das Paper mit meinem Eintrag bei pos: ra und dec. Nicht dass ich die beiden vertauscht habe.

### DIFF
--- a/input/data/2009/2009A%26A...499..723A/tev-000132.yaml
+++ b/input/data/2009/2009A%26A...499..723A/tev-000132.yaml
@@ -1,4 +1,26 @@
 source_id: 132
 reference_id: 2009A&A...499..723A
 
-# TODO: add data
+data:
+  # TODO: No livetime found in the paper
+  livetime:
+  significance: 12.1
+  excess: 689
+
+pos:
+  ra: {val: 40d23m9.2s, err: 2.4m, err_sys: 20s}
+  dec: {val: 0d47m10.1s, err: 2.4m, err_sys: 20s}
+
+morph:
+  type: gauss
+
+  # TODO: sigma and sigma2 have assymmetric errors: -0.03, +0.04, no 'pa' value found
+  sigma: {val: 0.34, err: 0.04}
+  sigma2: {val: 0.34, err: 0.03}
+
+spec:
+  type: pl
+
+  norm: {val: 4.14, err: 0.32, err_sys: 0.83}
+  index: {val: 2.1, err: 0.07, err_sys: 0.2}
+  ref: 1
--- a/input/data/2009/2009A%26A...499..723A/tev-000132.yaml
+++ b/input/data/2009/2009A%26A...499..723A/tev-000132.yaml
@@ -8,8 +8,8 @@ data:
   excess: 689
 
 pos:
-  glon: {val: 40d23m9.2s, err: 2.4m, err_sys: 20s}
-  glat: {val: -0d47m10.1s, err: 2.4m, err_sys: 20s}
+  glon: {val: 40d23m9.2s, err: 0d2.4m, err_sys: 0d20s}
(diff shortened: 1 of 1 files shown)


input/data/2009/2009A%26A...499..723A/tev-000132.yaml
```diff
@@ -1,4 +1,26 @@
 source_id: 132
 reference_id: 2009A&A...499..723A
 
-# TODO: add data
+data:
+  # TODO: No livetime found in the paper
+  livetime:
```
cdeil: livetime is 27 hours.

Section 2:
> The total dead-time corrected, quality and offset selected (2.5 ) data set amounts to 27 hours

input/data/2009/2009A%26A...499..723A/tev-000132.yaml
```diff
@@ -1,4 +1,26 @@
...
+  significance: 12.1
+  excess: 689
+
+pos:
+  ra: {val: 40d23m9.2s, err: 2.4m, err_sys: 20s}
+  dec: {val: 0d47m10.1s, err: 2.4m, err_sys: 20s}
```
cdeil: Missing minus sign: `val: -0d47m10.1s`

input/data/2009/2009A%26A...499..723A/tev-000132.yaml
```diff
@@ -1,4 +1,26 @@
...
+
+morph:
+  type: gauss
+
+  # TODO: sigma and sigma2 have assymmetric errors: -0.03, +0.04, no 'pa' value found
+  sigma: {val: 0.34, err: 0.04}
+  sigma2: {val: 0.34, err: 0.03}
```
cdeil: Remove sigma2. A symmetric, not elongated Gaussian was fitted.

input/data/2009/2009A%26A...499..723A/tev-000132.yaml
```diff
@@ -1,4 +1,26 @@
...
+  excess: 689
+
+pos:
+  ra: {val: 40d23m9.2s, err: 2.4m, err_sys: 20s}
+  dec: {val: 0d47m10.1s, err: 2.4m, err_sys: 20s}
+
+morph:
+  type: gauss
+
+  # TODO: sigma and sigma2 have assymmetric errors: -0.03, +0.04, no 'pa' value found
```
cdeil: Remove TODO comment.
You can keep a comment that the error given in the paper is asymmetrical, and that the down error is 0.03 deg.

input/data/2009/2009A%26A...499..723A/tev-000132.yaml
cdeil: In this case the position is given in the paper as l / b, i.e. in Galactic coordinates. (not in RA / DEC).
As mentioned at https://github.com/gammapy/gamma-cat/blob/master/input/schemas/dataset_source_info.schema.yaml#L45 for those cases we should use the key names `glon:` and `glat:`

input/data/2009/2009A%26A...499..723A/tev-000132.yaml
pdeiml: What about the "pa" value which I didn't found. Does it have to be given or is it optional?
If it has to be given, then we should think about some default value.

cdeil: Usually when values aren't given, the idea is to just not put the key at all in the input files.
So no default needed for `pa` and `sigma2` here, just omit it completely in the input file.

OK?

pdeiml:
Rechecked the errors.
Ready to merge.

cdeil:
Thanks!

PR #82, opened 2017-02-08
Always put unit for angular extensions

cdeil:
This is a suggestion to make the input data files more clear and less error-prone.

Currently for the source extensions `sigma` and `sigma2` we have a mix of "deg" and "arcmin", and sometimes no unit explicitly put down, produced with
```
ack sigma input/data
```
https://gist.github.com/cdeil/3bdd807b2b3fe2d0a0280b9d90361389

Currently, our schema is this:
https://github.com/gammapy/gamma-cat/blob/master/input/schemas/dataset_source_info.schema.yaml#L66
```
  morph:
    type: object
    additionalProperties: false
    properties:
      type: {enum: [point, gauss, shell]}
      sigma:
        type: object
        additionalProperties: false
        description: |
          Width parameter (semantics type-dependent)
          Number in deg or string that can be parsed by Angle
        properties:
          val: {type: [string, number]}
          err: {type: [string, number], description: Statistical error}
      sigma2:
        type: object
        additionalProperties: false
        description: |
          Width parameter (semantics type-dependent)
          Number in deg or string that can be parsed by Angle
        properties:
          val: {type: [string, number]}
          err: {type: [string, number], description: Statistical error}
```

I think e.g. the first source listed is incorrect, it's probably 6 arcmin in size, and not 6 deg:
```
$ ack sigma input/data
input/data/2005/2005A%26A...435L..17A/tev-000079.yaml
15:  sigma: {val: 6.4, err: 0.7}
16:  sigma2: {val: 2.3, err: 0.5}
```

---

My suggestion here would be to always do data entry with an angle, using this format:
```
In [1]: from astropy.coordinates import Angle

In [2]: Angle('3d')
Out[2]: <Angle 3.0 deg>

In [3]: Angle('3m')
Out[3]: <Angle 3.0 arcmin>
```
I.e. if the paper gives the value in "deg", put it in "deg", and if it's in "arcmin", put it in "arcmin".

The schema can be changed to only take string as input, and the description changed to say that format should be "<val>s" and "<val>m".

---

I think the number of cases where one needs to look at the paper again to figure out what the units are or double-check is small. This could be done as one PR, I think that will be OK to review and get in quickly.

@pdeiml - You or me?

(I won't have time this week, but could try to get it done next week.)

### DIFF
--- a/input/data/2005/2005A%26A...432L..25A/tev-000110.yaml
+++ b/input/data/2005/2005A%26A...432L..25A/tev-000110.yaml
@@ -6,8 +6,8 @@ data:
   significance: 13
 
 pos:
-  glon: {val: 0.872 deg, err: 0.005 deg}
-  glat: {val: 0.076 deg, err: 0.005 deg}
+  glon: {val: 0.872d, err: 0.005d}
+  glat: {val: 0.076d, err: 0.005d}
 
 morph:
   type: point
@@ -15,10 +15,8 @@ morph:
 
 spec:
   type: pl2
-
   norm: {val: 5.7, err: 0.7, err_sys: 1.2}
   index: {val: 2.40, err: 0.11, err_sys: 0.20}
   ref: 0.2
-
-  theta: 0.1
+  theta: 0.1d
   erange: {min: 0.17}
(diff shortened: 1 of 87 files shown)


cdeil:
I wanted to fix this. Commit attached. Merging now.

Angles in input files should be correct and uniform now.

PR #84, opened 2017-02-09
Data entry for paper 2015Sci...347..406H, source 26

pdeiml:
Selbiges Problem bzgl significance etc wie bei PR 83


pdeiml:
Ready to merge.

PR #105, opened 2017-04-23
Completed all HESS J0632+057 data

GernotMaier:
All published data on HESS J0632+057 from HESS/MAGIC/VERITAS is now in the catalogue


cdeil:
Thanks!

Issue #86, opened 2017-02-09
Remove source_name key in input files


cdeil:
Suggestion by @pdeiml - At the moment we sometimes put `source_name` in input files, and sometimes not.  The idea is that only `source_id` is used, and than `source_name` or `common_name` is only added to help with not getting confused about which file is for which source.

But putting names is problematic, it leads to inconsistencies between names used.
Options are:

a. Remove all `source_name` or `common_name` entries in input files.
b. Keep `common_name`, but document that it's optional and for convenience.

I'd like to sleep on it, but at the moment I'm leaning towards option a.

pdeiml:
Just for completeness:
We decided to use option a.

adonath:
@cdeil @pdeiml Where else then is the information on `common_name` and other source names etc. stored?

cdeil:
The basic source info for a given `source_id` is in the files in `input/sources`.
So the idea is to only keep `source_id` in the files in `input/data`, and then only have names and such in one location, the files in `input/sources`.

One thing I do think we should allow is to have comments in the input files, e.g. in cases where there's multiple sources in one paper, and it helps with data input and review to have comments explaining which source data is entered here.

adonath:
@cdeil Thanks for the clarification. :+1: for this change!

cdeil:
It's quick to find all cases via `ack 'source' input/data` and do the clean-up.

@pdeiml - You or me?

pdeiml:
@cdeil I can do it. Either today in the afternoon or on Monday, I will have time for that.
I think it is not an urgent thing.

cdeil:
> @cdeil I can do it. Either today in the afternoon or on Monday, I will have time for that.
I think it is not an urgent thing.

Thanks! Not urgent at all.
I just wanted to avoid you and me starting the same thing.
For this, let's use the "assignee" field on Github, i.e. assign issues to the one planning to take care of it.
So in this case, I'm assigning it to you, and you can just do it whenever you have time.

cdeil:
Implemented by @pdeiml in #95.
Closing now.

PR #110, opened 2017-04-24
two LS I +61 303 papers, CTA 1 SED

GernotMaier:


input/data/2008/2008ApJ...679.1427A/tev-000014-sed.ecsv
```diff
@@ -0,0 +1,24 @@
...
+# datatype:
+# - {name: e_min, unit: TeV, datatype: float64}
+# - {name: e_max, unit: TeV, datatype: float64}
+# - {name: dnde, unit: TeV-1 cm-2 s-1, datatype: float32}
+# - {name: dnde_err, unit: TeV-1 cm-2 s-1, datatype: float32}
+# - {name: dnde_ul, unit: TeV-1 cm-2 s-1, datatype: float32}
+# meta: !!omap
+# - data_type: sed
+# - source_id: 14
+# - reference_id: 2016ApJ...817L...7A
```
cdeil: This `reference_id: 2016ApJ...817L...7A` doesn't match the folder name: `2008ApJ...679.1427A/tev-000014-sed.ecsv`

input/data/2008/2008ApJ...679.1427A/tev-000014-sed.ecsv
```diff
@@ -0,0 +1,24 @@
+# %ECSV 0.9
+# ---
+# datatype:
+# - {name: e_min, unit: TeV, datatype: float64}
```
cdeil: So far, whenever we have a published differential spectrum, i.e. `dnde`, the energy `e_ref` at which the diff flux is `dnde` was given. Is this not the case here?

input/data/2008/2008ApJ...679.1427A/tev-000014.yaml
```diff
@@ -0,0 +1,35 @@
+# Source ID from TeGeV-cat: http://www.asdc.asi.it/tgevcat/)
+# LS I +61 303
+# VER J0240+612
+source_id: 14
+# Reference ID from ADS
```
cdeil: Remove comment "Reference ID from ADS". We have this almost everywhere, no need to comment.

input/data/2008/2008ApJ...679.1427A/tev-000014.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+#############################################################
+# spectral fit results
+# (spectral fit for phases 0.5-0.8)
+
+spec:
+  type: pl
+  mjd: {min: 53999.4, max: 54147.1}
+  phase: {min: 0.5, max: 0.8 }
+  norm: {val: 2.89e-12, err: 0.32e-12, err_sys: 0.6e-12}
```
cdeil: Remove the `1e-12` part here when doing the data entry, i.e. change to this:
```
norm: {val: 2.89, err: 0.32, err_sys: 0.6}
```

We might change this in the future, but for now data entry has to be done this way.

The 1e-12 is noted as a scale with the unit in the spec:
https://github.com/gammapy/gamma-cat/blob/master/input/schemas/dataset_source_info.schema.yaml#L190

input/data/2011/2011ApJ...738....3A/tev-000014-lc.ecsv
```diff
@@ -0,0 +1,65 @@
...
+# - {name: e_min, unit: TeV, datatype: float64}
+# - {name: time, unit: MJD, datatype: float64}
+# - {name: significance, datatype: float32}
+# - {name: flux, unit: ph / (cm2 s), datatype: float64}
+# - {name: flux_err, unit: ph / (cm2 s), datatype: float64}
+# - {name: flux_ul, unit: ph / (cm2 s), datatype: float64}
+# meta: !!omap
+# - data_type: lc
+# - source_id: 14
+# - source_name: LS I +61 303
```
cdeil: We've decided to remove `source_name` keys from data input files, and to only keep `source_id`.
So please remove it here.

The reasoning (see https://github.com/gammapy/gamma-cat/issues/86) was that it's extra work to check two pieced of info that contain the same info.

While I'm doing data entry, I'm using
https://gammapy.github.io/gamma-cat/sources.html
to look up `source_id` from names.

input/data/2008/2008ApJ...679.1427A/tev-000014-sed.ecsv
GernotMaier: OK - wasn't aware of that. I've added the e_ref column.

input/data/2008/2008ApJ...679.1427A/tev-000014-sed.ecsv
GernotMaier: Fixed.

input/data/2008/2008ApJ...679.1427A/tev-000014.yaml
GernotMaier: OK - I usually don't delete comment. Maybe someone is just looking at one file from one source, then any help might be appreciated.

input/data/2008/2008ApJ...679.1427A/tev-000014.yaml
GernotMaier: OK - didn't know. I would prefer to write explicitly everywhere the units. It makes it more transparent and doesn't require any insider knowledge.

Fixed this.

input/data/2011/2011ApJ...738....3A/tev-000014-lc.ecsv
GernotMaier: OK - I've fixed this.

input/data/2008/2008ApJ...679.1427A/tev-000014-sed.ecsv
GernotMaier: Coming back to this: there are several publication where e_ref is not given, but differential fluxes are quoted in tables between e_min and e_max. Can't we add flexibility that in this case, e_ref is calculated dynamically? My feeling is that we add here information to the catalogue, which is not part of the publication.

Would have to discuss of course what to do:
- simple mean value
- weighted mean (with weights calculated from shape of the spectrum)
- ...

cdeil: Well, to me whatever we do is all pretty equally bad: at some point we're inventing an `e_ref` at which to put the published `dnde`. I think if I were to do it, I would add it here in the dataset and put a caveat as a comment that this `e_ref` was made up, not published.

If you put it in the tools (e.g. Gammapy SED reader) you're just encouraging bad behaviour to publish a measurement in a bad way. (And it is rare, I haven't seen that before and we have SEDs from > 100 papers by now).

cdeil: It's this spectrum, no?
http://iopscience.iop.org/article/10.1086/587736/fulltext/22376.figures.html
For such a spectrum (simple shape, bins not very wide) results don't depend strongly on what you do.
Would suggest to compute `e_ref = sqrt(e_min * e_max)`, i.e. log bin center here and add that as a column, and a one-line comment in the header saying that this is what was done.

GernotMaier: Yes, it is this paper. I have a few more on which I am working on. I can do this, and add a comment. But I still think this is not the correct solution - I don't think it is a bad way to let the tools calculate this later.

(actually where e_ref should be is not straightforward, and mostly not mentioned in publications).

cdeil: I see that the paper doesn't give a table with the spectral point information (like most papers). So it's not like in this case `e_min` or `e_max` or `e_ref` or `flux` or `dnde` or anything was really "published". The spectral points shown in Figure 2 are `dnde` drawn at some `e_ref` values.

Irrespective of the question if you add the `e_ref` column or not (we'll merge either way): can you add a "comment" or "url" in the header where this data is coming from?
If you look at other examples you'll see that it's always either a "url" pointing to a webpage or file with the information, or a "comment" saying something about the origin of the data (in this case I presume a VERITAS-internal webpage or from corresponding author).

GernotMaier: The corresponding author was a certain G.Maier and the values are from his notes... I've added a comment.

cdeil:
Are you still working on this PR or is it ready to be merged?
(in case it's ready: please always leave a comment, so that I know when to look at it)

GernotMaier:
It is ready to be merged - I addressed all of the issues.

cdeil:
I've done a little bit of cleanup in master. Removed "ph" from units, I decided to consistently not use those in gamma-cat, otherwise all looked good!

PR #119, opened 2017-05-11
Change spectral model input format

cdeil:
I'm changing the spectral model input format in this pull request:

- [x] Adapt all input YAML files
- [x] Adapt the schema
- [x] Adapt the scripts

The goal is to get rid of the data entry issues we had for PL2 and ECPL models by having a more clear / explicit data entry scheme.

This is possibly a format we could support in Gammapy via the generic `to_dict` / `from_dict` methods (but I don't have time to work on that today).


cdeil:
I'm merging this now. And I'll do a follow-up PR changing the spectral model parameter output format in the catalog.

The diff here is reasonable, although there's ~ 7 PL2 spectra where some values in the output catalog were lost, because of emin / emax handling. The previous results were possibly incorrect anyways, and this should be fixed later today when the catalog output format is changed.
We can use spectral plots to compare against flux points to verify.

PR #120, opened 2017-05-12
Add of Biteau catalog data

pdeiml:
The following sources have not been added due to some issues:

- [x] Mkn421 - 2002A&A...393...89A:
There are two parts in the catalog with same energy but different fluxes and start/end times. 
How do we want to compare these two measurements?

- [x] Mkn421 - 2005A&A...437...95A:
There is already a folder and a .ecsv file in there with same measurement times?

- [x] Mkn421 - 2012JPhG...39d5201C:
There are two measurements with different timestamps but no "note" is given.
How do we want to compare these two measurements?

- [x] PKS 2155-304 - 2005A&A..430..865A:
Obviously, in this reference, there are more than one measurement but no "note" is given.
How do we want to compare these two measurements?

- [x] PKS 1222+21 - 2011ApJ...730L...8A:
In /input/sources, there is no source with this name, only source number 56 is PKS 1222+216.
Is this the same source?

- [x] H 1426+428 - 2003A&A...403..523A:
There are two measurements with different timestamps but no "note" is given.
How do we want to compare these two measurements?

- [x] H 2356-309 - 2010A&A...516A..56H:
There are two measurements with different timestamps but no "note" is given.
How do we want to compare these two measurements?

- [x] PKS 1424+240 - 2014ApJ...78L..16A:
There are two measurements with different timestamps but no "note" is given.
How do we want to compare these two measurements?


cdeil:
Is this a PR that you want to be merged when ready?
If yes, please put a better title and description (what you are adding).

Let me know if you have any issue where you need my help.

pdeiml:
Yes, it is a pull request that should be merged. I thought the title of the PR is equal to the name of the branch :-)
I will change it.

Indeed, I need your help. We have to discuss about the issues above that I wrote down as reminders for myself.

pdeiml:
Hope that the title is now much better

cdeil:
@pdeiml - Change the comment above to use task list that you can tick off as you work towards finalising this issue (and that I can look at for status / where to suggest help)?
https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments

---

> Mkn421 - 2005A&A...437...95A:
There is already a folder and a .ecsv file in there with same measurement times?

You have to check if we already have that data entered. If yes, you have to add an `if` to your script to skip that dataset

---- 

> PKS 1222+21 - 2011ApJ...730L...8A:
In /input/sources, there is no source with this name, only source number 56 is PKS 1222+216.
Is this the same source?

Yes, this is the same source. See e.g. here:
http://simbad.u-strasbg.fr/simbad/sim-basic?Ident=PKS+1222%2B21+&submit=SIMBAD+search#lab_ident

other_data_collections/2015ApJ...812...60B/biteau.py
```diff
@@ -1,46 +1,177 @@
...
-        # import IPython; IPython.embed()
-
-
-if __name__ == '__main__':
-    maker = BiteauMaker()
-    # biteau.run_basic_checks()
-    maker.run_checks()
+    for i in range(0, len(table)):
+        if((table['source'][i]==source) and (table['reference_id'][i] == reference_id) and (table['experiment'][i] == experiment)):
+            new_table.add_row((4.135667662E-27*table[i]['freq'], table[i]['e2dnde']/1.6022, table[i]['e2dnde_errn']/1.6022, table[i]['e2dnde_errp']/1.6022))
```
cdeil: This is not well readable.
Please use `astropy.units` or at least given your magic constants like 4.1e-27 and 1.6 names.
And wrap the line so that it's ~ 80 char max width, so that the code becomes readable without having to scroll back and forth.
Maybe a `row = table[i]` would also help (instead of using `[i]` everywhere in the formula that does the flux computation.

cdeil:
Why put the ~ 10 lines in `other_data_collections/2015ApJ...812...60B/add_biteau.py` into a separate script. Please keep everything in the one `biteau.py`, and put this code into a function that you can execute at the bottom in the `if __name__ == '__main__'` section with one line.

cdeil:
All of these SEDs should have a one-line comment where the data comes from.
This is important to give credit, but also because we didn't check the data ourselves and there might be issues.
E.g. I seem to remember that Biteau didn't include upper limits, so in this sense the datasets entered here aren't the complete info given in the original papers.

pdeiml:
[ ] Test

cdeil:
@pdeiml - Thanks for this big pull request!

Comments:

* Please rebase or merge this PR against the latest `master`. Specifically, this PR should not have touched the unrelated files in `2014ApJ...780..168A` and `2016arXiv161003751S` where you re-applied the `filecounter` -> `file_id` change that was already done in master a while ago.
* In the task list in your last comment you list `Change "telescope" to "telescope_id"`. This change should not be done, we use "telescope".
* I'm still not happy with the code where you compute the fluxes. It's not clear to me looking at the code if it's correct or not. My understanding is that you have no way to re-run the script anyways because you would loose many manual edits? If yes, then I could check on one or two cases if the fluxes are correct.
* The comment line `This data was collected for 2000AIPC..515..113P and contributed to gamma-cat by Janathan Biteau.` should be changed to: `This data was collected for 2015ApJ...812...60B and contributed to gamma-cat by Jonathan Biteau.` (see https://github.com/gammapy/gamma-cat/tree/master/other_data_collections/2015ApJ...812...60B#agn-spectra-from-biteau--williams-2015 and note the typo fix "Janathan" -> "Jonathan".
* There's a lot of "note" keys which you copied from the column of that name by Jonathan, e.g. `{note: '2006-04-22'}` or `{note: Flux4}`. We don't have that key in gamma-cat, and I don't think it's something we want to add / maintain. I think in some cases (like Flux4) this is just equivalent to the `file_id` we have, doesn't add any useful info and can be removed. Similarly for the dates, I think that info is always contained already in `mjd` keys. So I think all of these "note" keys can / should just be deleted. Alternatively, if you think they are of value, you could reformat them to be an extra comment line saying: "The Biteau table contained the following note: Flux4".

pdeiml:
- [ ] Rename the comments that the data are from Biteau
- [ ] Remove the keyword "note"
- [ ] Add mjd as keyword

cdeil:
@pdeiml - There's another merge conflict now in `gammacat/sed.py`. Please resolve.

I've done a check for the `dnde` fluxes using one SED as example:
https://gist.github.com/cdeil/2942d2f0241d7ff8d25db8008e0fda03
@pdeiml - A bug in your script or misunderstanding what `dnde` is?
Or did I make a mistake / take the wrong input file or columns from the Biteau table with all the SEDs?

Usually AGN SEDs are steep power-laws with spectral index ~ 3. So going up one decade in energy (from e_ref = 1 to 10 TeV), the expectation is that the differential flux `dnde` falls by 3 decades. So the `dnde` I get are consistent with that expectation, your `dnde` values at high energies are much too high to be correct (I'm pretty sure, but not 100%).

cdeil:
@pdeiml - I think you've started over and are implementing this now in #129 ?
If yes, please close this PR.

pdeiml:
I close this PR because the data entry of Biteau's catalog is now in PR 129

PR #121, opened 2017-05-12
Data entry for some Galactic sources

cdeil:
This PR does some data entry for Galactic sources.
The format is already as in #119 , I'll now try to finish up that other one first before merging this one.


cdeil:
Merging this now. I'll push the changes through to the output catalog later today.

PR #122, opened 2017-05-12
Better spectral model parameter format in output catalog

cdeil:
This is a follow-up to #119: change the gamma-cat catalog format to have more explicit columns for the spectral mode parameters.

Discussed today with @adonath . This will require corresponding changes in gammapy.catalog (which I'll make now) and HGPS (which @adonath agreed to make).


cdeil:
I'm merging this now. The diff is hard to review, and it's easiest to push the change through to gammapy and then look at plots.

PR #137, opened 2017-07-19
more VERITAS data

GernotMaier:
This pull request contains:

- two additional VERITAS papers
- 2016ApJ...821..129A (Galactic ridge) required to introduce a new source ID which is not in TeGeV (ID 165)
- broken power law model results for Sgr A* are given in comments
- Tycho spectral results correction (UL value and reference energy)


cdeil:
@GernotMaier - Thanks!

I've attached a few commits here with small fixes.

cdeil:
Output files updated in 38a2626 .

Issue #133, opened 2017-07-13
pip install regions fail on travis-ci


cdeil:
The gamma-cat build is failing on travis-ci and RTD now:
* https://travis-ci.org/gammapy/gamma-cat/builds/253190303#L365
* https://readthedocs.org/projects/gamma-cat/builds/5683433/

@bsipicz - Do you understand why? I think this is an issue that was fixed in astropy-helpers, but it still shows up in the older `regions` 0.2 release? Does that mean we need to do a bugfix release for `regions` and possibly other packages?

cdeil:
@pdeiml ran into the same issue and reported it via email.

This was discussed at length in https://github.com/sphinx-doc/sphinx/issues/3766 and https://github.com/astropy/astropy-helpers/issues/302#issuecomment-305976972 but I don't remember how it was fixed. This now works for me with Python 3.6 the existing regions 0.2 release and Sphinx 1.6.3 without triggering the TypeError:
````
python -m venv spam
source spam/bin/activate
pip install sphinx
pip install regions
```

@bsipocz - Can you still reproduce the issue or have advice for @pdeiml that is facing it? Do we need to make a new `regions` release that contains the `astropy-helpers` fix?

pdeiml:
Hello,
as @cdeil mentioned I ran into the same problem. But after full deletion of everything (gammapy, gamma-cat and gamma-cat conda environment) and reinstallation everything worked fine. I don't know what the problem is/ was.

cdeil:
@pdeiml - Thanks for the feedback.

Let's see if the issue is still present on travis-ci and RTD (see #139) and if not I'll close this issue.

cdeil:
I did manage to fix the issue by changing to `conda install regions` (or work around, not sure what the core of the issue was with the `pip install regions`):
* https://github.com/gammapy/gamma-cat/commit/0acbf8f1adf9c9b1d914ba8f44b1d6d258768d10
* https://travis-ci.org/gammapy/gamma-cat/builds/267476552

cdeil:
After a few more tweaks (directly in `master`), the RTD build for gamma-cat works again:
* https://readthedocs.org/projects/gamma-cat/builds/5872498/
* (e.g. this new docs entry appears: http://gamma-cat.readthedocs.io/contribute/code.html#installation ).

Makes me 😄 .

PR #139, opened 2017-08-16
Update Python 3.5 -> 3.6

cdeil:
This PR updates the build on travis-ci and RTD from Python 3.5 to 3.6.

It shouldn't really matter, but I wanted to see if the issue from #133 is still present on travis-ci, and it's easier to debug if it's the same version there that I use locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ script:
 # Workaround for the following issue:
 # https://travis-ci.org/gammapy/gamma-cat/builds/225367662#L755
 # Reported here: https://github.com/bokeh/bokeh/issues/6193
-- export BOKEH_DOCS_MISSING_API_KEY_OK=yes
+#- export BOKEH_DOCS_MISSING_API_KEY_OK=yes
 
 - export PYTHONPATH=.:$PYTHONPATH
 - python make.py --help
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@ An open data collection and source catalog for gamma-ray astronomy.
 ## See also
 
 * http://gamma-sky.net/
-* http://gamma-sky.net/cat/tev
 * http://docs.gammapy.org/en/latest/catalog/index.html
 
 ## Infos for maintainers and contributors
--- a/documentation/contribute/code.rst
+++ b/documentation/contribute/code.rst
@@ -9,6 +9,21 @@ It's a bunch of Python modules in the ``gammacat`` folder.
 This is only useful if you want to contribute to the gamma-cat.
 Users can use ``gammacat.catalog`` to access the catalog information.
 
+Installation
+------------
+
+If you want to run the ``gamma-cat`` Python scripts locally, you need to install
+Python 3.6 and some Python packages. We recommend you download `Anaconda <https://www.continuum.io/downloads>`__
+and then install the packages via::
+
+    conda env create -f environment.yml
+    source activate gamma-cat
+
+We have continuous integration tests set up on travis-ci that check that everything is working OK.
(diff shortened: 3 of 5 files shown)


cdeil:
The `pip install regions` still fails, now with this error:
https://travis-ci.org/gammapy/gamma-cat/builds/265038022#L444

@bsipocz - Any advice what to do here?

pdeiml:
@cdeil: can you change line 9 in environment.yml from 'python=3.5' to 'python=3.6' and see what happens?

cdeil:
> can you change line 9 in environment.yml from 'python=3.5' to 'python=3.6' and see what happens?

@pdeiml - I already did that in this PR: https://github.com/gammapy/gamma-cat/pull/139/files#diff-b7cb75ca59993a879e29bed8cd8469b8R9

pdeiml:
Ok, sorry, I accidentally checked the version in the master branch and not in this PR.

I did the following on a clean Ubuntu 16.04 PC with anaconda3 on it:

Firstly, I cloned my forked gamma-cat/gammapy repos (master is uptodate):
**mkdir gammapy gamma-cat
cd gammapy
git clone https://github.com/pdeiml/gamma-cat
cd ../gammapy
git clone https://github.com/pdeiml/gammapy**

Then in gamma-cat folder:
**conda config --set always_yes yes --set changeps1 no
condo update -q condo
conda info -a
conda env create -f environment.yml** 

This leads to the same error as you have with travisCI (note that my paths with the errors refer to are something like this: anaconda3/envs/gamma-cat/lib/python3.5.)

I thought that maybe there is something wrong with gammapy, hence, next step was to install gammapy in the following way (in the gammapy folder):
**python setup.py install --user**
which works fine and gammapy is installed into /.local/lib/python3.6/site-packages. This means that python 3.6 is used.

After that I deleted the gamma-cat environment and executed again:
**conda env create -f environment.yml**
This still leads to the same errors as yours and again with python3.5 in the paths of the errors.

then I changes python 3.5 to 3.6. in environment.yml, deleted the gamma-cat environment and re-executed
**conda env create -f environment.yml**

and that works!

cdeil:
@pdeiml - Thanks for the info!

But we're not going to do any install / uninstall dances on travis-ci like you did.
It's OK to leave CI broken here for now, until proper fixes are done (possibly a new release of the `regions` package which is the culprit here, because it's the `pip install regions` step that fails).

pdeiml:
Ok, the solution is definitely my installation of gammapy with 'python setup.py install --user'.

I cleaned everything, changed python3.5 to python 3.6 without installing gammapy and then still the errors occur.

Hence, the upper guide is the solution how to install gamma-cat and gammapy on local machines but it is not a solution for the travisCI problem.

@cdeil: should we put the upper guide on https://gamma-cat.readthedocs.io?

cdeil:
@pdeiml - Yes, we should have install instructions as a small section above this section in the contributor docs:
https://gamma-cat.readthedocs.io/contribute/code.html#make-py

I would like it to be very short though and to not contain any workarounds.
If there's conda or pip install issues, we need to fix them anyways.

My suggestion would be to simply put one sentence: if you want to run the gamma-cat scripts locally, use Anaconda Python and copy & paste the install commands from https://github.com/gammapy/gamma-cat/blob/master/.travis.yml

@pdeiml - Want to make a PR to update the contributor docs?

cdeil:
@pdeiml - I'm merging this PR now because it contains the update to Python 3.6 and I added a section with install instructions to the docs  (#133 is still open as a reminder about the `pip install regions` issue).

PR #140, opened 2017-08-16
Addition of more VERITAS data

GernotMaier:
This pull request is a continuation of the work on adding all VERITAS data:

- contains a few more papers (all additions are always the complete data set for the corresponding paper)
- had to add a few source IDs, as not all TeV sources are in the TeGeV catalogue

(And I apologize that I didn't find time to install gammpy to run the 'make.py check' - hope this is ok)


cdeil:
@GernotMaier - Is the errp / errn change in input/data/2011/2011ApJ...730L..20A/tev-000003-sed.ecsv correct or is it switched now? (the order of columns in the ECSV header and the CSV header line has to match to be a valid ECSV file)

---

I see you're adding the Crab pulsar as a separate source from the Crab nebula. We haven't discussed this, I'm 50:50 on the question whether to split that into two "sources" or have one source with multiple "components". TeVCat has two IDs for it, 3FGL has PSR and PWN as separate sources, 3FHL only has one source (with class PSR). :-)
So OK, let's split out TeV pulsars (starting with Crab) as separate sources.

---

For the new source `SNR G106.3+02.7` that you're adding in `input/sources/tev-000149.yaml`: is it clearly a different source from the Boomerang PWN that we already have in gamma-cat? I.e. is it clearly a good choice to put it as a new separate source? I see TeVCat also has it as a separate source, so probably the way to go?

* https://github.com/gammapy/gamma-cat/blob/master/input/sources/tev-000150.yaml
* http://gamma-sky.net/#/cat/tev/150

---

Apart from these questions, I think this is ready to be merged, no?

GernotMaier:
>  Is the errp / errn change in input/data/2011/2011ApJ...730L..20A/tev-000003-sed.ecsv correct or is it switched now? (the order of columns in the ECSV header and the CSV header line has to match to be a valid ECSV file)

Yes - it is first errn and then errp (correct now).

> I see you're adding the Crab pulsar as a separate source from the Crab nebula. We haven't discussed this, I'm 50:50 on the question whether to split that into two "sources" or have one source with multiple "components". TeVCat has two IDs for it, 3FGL has PSR and PWN as separate sources, 3FHL only has one source (with class PSR). :-)
So OK, let's split out TeV pulsars (starting with Crab) as separate sources.

I think these are different sources with different emission mechanism, so separate them is logical for me.

> For the new source SNR G106.3+02.7 that you're adding in input/sources/tev-000149.yaml: is it clearly a different source from the Boomerang PWN that we already have in gamma-cat? I.e. is it clearly a good choice to put it as a new separate source? I see TeVCat also has it as a separate source, so probably the way to go?

I think these might be different sources, see the attached Figure 1 from the paper. The open yellow cross is at the pulsar position and you can see the Boomerang-shaped radio contour around it (only a little bit larger than the yellow cross).

screen shot 2017-08-16 at 16 31 14

Yes - we are ready to merge from my side.

cdeil:
I reviewed the diff and all values look reasonable.
Made one more small change, putting a more precise position for the Crab pulsar from SIMBAD:
```
In [1]: from gammacat.utils import print_simbad_pos

In [2]: print_simbad_pos('V* CM Tau')
pos:
  simbad_id: V* CM Tau
  ra: 83.63307625
  dec: 22.014493278
```
and changing the "discovery date" for the Crab pulsar to the value that's in TeVCat (`2008-11`)
(discovery date is supposed to be the discovery date in TeV, and if we say the Crab pulsar is separate from the nebula, then at TeV is was discovered in 2008, not 1989 like the nebula.

cdeil:
Thanks for answering my questions. Merging now.

PR #141, opened 2017-08-17
Add unit to livetime

pdeiml:
make.py checks 
runs without errors, hence the changes should be fine.
Do you know a better commit description?

### DIFF
--- a/input/data/2013/2013ApJ...764...38A/tev-000001.yaml
+++ b/input/data/2013/2013ApJ...764...38A/tev-000001.yaml
@@ -3,7 +3,7 @@ reference_id: 2013ApJ...764...38A
 telescope: veritas
 
 data:
-  livetime: 41.25
+  livetime: 41.25h
   significance: 7.5
   significance_post_trial: 6.5
 
--- a/gammacat/catalog.py
+++ b/gammacat/catalog.py
@@ -245,7 +245,8 @@ class CatalogSource:
             data['significance'] = NA.fill_value['number']
 
         try:
-            data['livetime'] = dsi['data']['livetime']
+            # data['livetime'] = dsi['data']['livetime']
+            data['livetime'] = Quantity(dsi['data']['livetime']).to('hour').value
         except KeyError:
             data['livetime'] = NA.fill_value['number']
 
--- a/input/data/2004/2004A%26A...425L..13A/tev-000106.yaml
+++ b/input/data/2004/2004A%26A...425L..13A/tev-000106.yaml
@@ -6,7 +6,7 @@ telescope: hess
 # called "model1" from July / August
 
 data: 
-  livetime: 11.8
+  livetime: 11.8h
   significance: 9.2
 
 pos:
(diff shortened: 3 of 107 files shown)


cdeil:
Can you also try `./make.py all` or `./make.py catalog`?

I think the line here:
https://github.com/gammapy/gamma-cat/blob/ecb4648cf94a6cb7f31cb6070763a6437d1f9d96/gammacat/catalog.py#L248
has to be changed to 
```
data['livetime'] = Quantity(dsi['data']['livetime']).to('hour').value
```
so that it works, no?

After this change, the output catalog shouldn't change with this PR.
This is something you can check via `git diff docs/data/gammacat.yaml` or `git diff docs/data/gammacat.ecsv`, to see what changed in the generated output catalog.

cdeil:
To check if unit handling for this input really works, you could also change one of the livetimes e.g. from "1 hour" to "60 min" and in the output files it should be "1 hour" in both cases.

pdeiml:
You were right. I changed line 248 in 
gamma-cat/gammacat/catalog.py. 
Now make.py all does work.

I checked the unit conversion by changing the lifetime of 41.25h in 2013/2013ApJ...764...38A/tev-000001.yaml
to 2475min and in both cases in gammacat.fits.gz the corresponding entry is 41.25.

So this works and it is ready to merge. For the other conversion, I think, a new PR is better than to put everything in this PR.

By the way: How did you do this nice link to a specific line in a file in you upper comment?

pdeiml:
The last commit was necessary to fix some merge conflicts :-)

cdeil:
This is not OK, I don't want to merge this:
https://github.com/gammapy/gamma-cat/pull/141/commits
It will lead to very complex version history and is complicated / time consuming to review.

I think we've discussed it before, but please: don't merge master into your feature branch; instead rebase your feature branch on master.

cdeil:
> By the way: How did you do this nice link to a specific line in a file in you upper comment?

I just linked to a line of code, and then Github inserted that for me that way:
https://github.com/blog/2415-introducing-embedded-code-snippets

pdeiml:
Ok, nice this thing with the linked lines :-)

Now this PR is ready to merge.

cdeil:
> Now this PR is ready to merge.

No. This PR shows commits from Gernot and me, because you merged master into this branch. Please see my comment on that matter above and fix.

pdeiml:
But now, it is ready to merge

input/data/2013/2013ApJ...764...38A/tev-000001.yaml
```diff
@@ -3,7 +3,7 @@ reference_id: 2013ApJ...764...38A
 telescope: veritas
 
 data:
-  livetime: 41.25
+  livetime: 2475min
```
cdeil: Please go back to hours here. Trying out minutes was just for testing.

pdeiml:
ah sry, that was from the unit testing. 
Fixed it

pdeiml:
Think it is ready to merge.

cdeil:
Looks good. Thanks!

PR #142, opened 2017-08-18
Add unit to position value in input files

pdeiml:
I think period is not used in the output files (grep does not find period in /docs/data). 
Anyway, I changed period type to string to have in the input files e.g.
period: {val: 345d}

python make.py all 
works.

Able to merge?!


cdeil:
Looks good to me. Thanks!

PR #144, opened 2017-08-21
Issue_112_zenith

pdeiml:
Zenith is not used in the output files, I think.
Hence, I cannot check wether different units are parsed correctly.

@cdeil do you think the description like
      livetime:
        type: string
        description: |
          Livetime (observation time minus dead)
          String that can be parsed by Astropy Units.

is enough or should we add an example at the end, like:
          String that can be parsed by Astropy Units (e.g. h, min,...)

Ready to merge.

### DIFF
--- a/input/data/2009/2009ApJ...698L.133A/tev-000029.yaml
+++ b/input/data/2009/2009ApJ...698L.133A/tev-000029.yaml
@@ -3,7 +3,7 @@ reference_id: 2009ApJ...698L.133A
 telescope: veritas
 
 data:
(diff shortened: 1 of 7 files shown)


cdeil:
Looks good to me. Merging.

> is enough or should we add an example at the end, like:
> String that can be parsed by Astropy Units (e.g. h, min,...)

I would suggest to keep it very short here, and then to explain that we use strings that can be parsed via Astropy Quantity in the docs with one example (but not duplicate that info in the description of every entry in the spec).

PR #145, opened 2017-08-21
issue_112_offset

pdeiml:
make.py all is working.
Think offset is not used in the output files. Hence, transformation of the units is not checked.

Ready to merge.


cdeil:
Thanks!

PR #146, opened 2017-08-21
Issue_112_morph_pa

pdeiml:
make.py all works.
I checked the unit transformation which is then written into gamma cat.fits.gz.

I rebased this PR against #145, as soon as #145 is merged to master, i need to do here a rebase against master.
Then ready to merge


cdeil:
Thanks!

PR #147, opened 2017-08-22
issue_112_erange

pdeiml:
@cdeil: 
Can you please check the code in catalog.py.
python make.py breaks exactly at 
https://github.com/pdeiml/gamma-cat/blob/issue_112_erange/gammacat/catalog.py#L670

The input of erange is done at
https://github.com/pdeiml/gamma-cat/blob/issue_112_erange/gammacat/catalog.py#L337

For completeness I added three output lines:
https://github.com/pdeiml/gamma-cat/blob/issue_112_erange/gammacat/catalog.py#L333
https://github.com/pdeiml/gamma-cat/blob/issue_112_erange/gammacat/catalog.py#L354
https://github.com/pdeiml/gamma-cat/blob/issue_112_erange/gammacat/catalog.py#L355

which show that e.g. the type of data['spec_ecpl_norm'] and of the erange values are 
astropy.Units.Quantity object.

I don't know why the part with erange breaks.

### DIFF
--- a/gammacat/catalog.py
+++ b/gammacat/catalog.py
@@ -330,29 +330,22 @@ class CatalogSource:
         else:
             raise ValueError('Unknown spectral model type: {}'.format(spec_type))
 
-        print(type(data['spec_ecpl_norm'])) #peter.deiml@fau.de
-
     @staticmethod
     def fill_spectral_other_info(data, dsi):
         try:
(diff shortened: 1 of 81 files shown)


pdeiml:
Should I rebase the two commits to one single commit?!

PR #148, opened 2017-08-22
issue_112_mjd0

pdeiml:
I think mjd0 is not used in output files, hence, parsing by astropy was not checked.

Ready to merge.


cdeil:
LGTM, merging now.

PR #149, opened 2017-08-23
issue_112_period

pdeiml:
I think the period is not used in the output files, hence, checking the unit transformation by astropy could not be checked.

python make.py all works without errors.

Ready to merge

### DIFF
--- a/input/schemas/dataset_source_info.schema.yaml
+++ b/input/schemas/dataset_source_info.schema.yaml
@@ -74,7 +74,7 @@ properties:
         additionalProperties: false
         description: |
           Right Ascension (ICRC or FK5 J2000, it's almost the same)
-          A string that can be parsed by Astropy Angle.
+          Properties are strings that can be parsed by Astropy Angle.
         properties:
           val: {type: string, description: Measured value}
           err: {type: string, description: Statistical error}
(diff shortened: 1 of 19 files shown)


pdeiml:
Do I understand you right? You want e.g. that 'glon' is like this:

      glon:
        type: object
       unit: deg
        additionalProperties: false
        description: |
          Galactic longitude 
          A quantity string with unit equal to 'degrees'.
        properties:
          val: {type: string, description: Measured value}
          err: {type: string, description: Statistical error}
          err_sys: {type: string, description: Systematic error}

So that everywhere 'unit: xxx' is mentioned?

cdeil:
Suggest to put this:
```
 glon:
    type: object
    additionalProperties: false
    description: |
      Galactic longitude 
      Properties are quantity strings with unit equivalent to "deg"
    properties:
      val: {type: string, description: Measured value}
      err: {type: string, description: Statistical error}
      err_sys: {type: string, description: Systematic error}
```

pdeiml:
Now, everything should be fine.
RTM

pdeiml:
Solved the conflicts, rebased against master.

RTM

input/schemas/dataset_source_info.schema.yaml
```diff
@@ -167,14 +167,16 @@ properties:
       period:
         type: object
         additionalProperties: false
-        unit: day
-        description: Orbital period
+        description: |
+          Orbital period
+          Property 'unit' is a string that can be parsed by Astropy Units.
```
cdeil: Suggest to remove this here:
```
Property 'unit' is a string that can be parsed by Astropy Units.
```
and to put this info below:
```
unit: {type: string, description: Unit equivalent with "day".}
```

pdeiml:
Changed it, 

RTM

cdeil:
Thanks!

PR #155, opened 2017-08-29
Add info.yaml files in input/data

pdeiml:
I added two keywords to dataset_source_info.schema.yaml:
data_entry_status: {missing, incomplete, complete}, required keyword
review_required: {yes, no}, optional keyword

For existing files I decided to add the keywords with the following logic:
- No TODO in the file: data_entry_status = complete, review_required: yes
- Some data and at least one TODO in the file: data_entry_status = incomplete
- Dummy file with no data (except reference_id, source_id): data_entry_status = missing

I would like to proceed as follows:
When adding new data to any yaml-file and you think it is perfect (hence, without TODO remarks), then write 'review_required: yes' and @cdeil all files that you have reviewed you can mark with 'review_required: no' to not get confused about what you already reviewed.

Hence, grepping 'incomplete' in /input/data will show all files with something to do in it (with TODO remarks), grepping 'missing' will show files without any kind of data, grepping 'review_required' will show files which need to be reviewed.

Issue #154, opened 2017-08-28
Keyword 'data entry status'


pdeiml:
This is an issue which I assigned to myself:

At the moment we have the problem that we cannot easily say whether gamma-cat is complete or not.

To make this easier, we should add the keyword

data_entry_status = {missing, incomplete, complete}

to all files in /input/data/.

In addition, make.py should get a function that prints a list of all sources and the data_entry_status.

cdeil:
@pdeiml - My suggestion would be to have general discussion on MAGIC in #7, and otherwise create one PR (or issue if useful) per MAGIC paper.

From experience: If you make one issue to discussion several papers like above, the discussion quickly gets very confusing and time-consuming, because it is a linear thread with multiple parallel discussions interspersed.

And concerning the papers you mention: I have no idea and no time to look. Just generally: yes, we're missing almost all MAGIC data.

---

If this issue is just your personal "notepad", then that's fine, I do that often as well to keep track, but then please clearly mention that when posting the issue and assign to yourself. (and I will unsubscribe from the issue notifications, to keep my inbox manageable)

cdeil:
Done by @pdeiml in #155 , continued in #159 .
I'm closing this issue now.

@pdeiml - please check `gamma-cat` master when you're back from vacation, and feel free to open further issues or PRs proposing further improvements.

PR #159, opened 2017-09-06
Info info.yaml files for input datasets

cdeil:
This PR is a continuation of #155 by @pdeiml (see also #154).
It adds an `info.yaml` file for each input dataset, as discussed in #155 and via Skype.

Task list:
- [x] Add schema for `info.yaml`
- [x] Revert some accidental edits to input files
- [x] Add schema validation for `info.yaml`
- [x] Review the diff


cdeil:
@pdeiml - FYI: One issue is that datasets is an array, not an object, i.e. dashes are missing everywhere, it should be this:
```
datasets:
  - tev-000106.yaml
  - tev-000106-sed.ecsv
```

cdeil:
Other changes I'm making to make `info.yaml` files:
- Fix typos or missing things (as I see them or pointed out via schema validation)
- Delete notes that don't add extra info, but just repeat the info from the "status" and "reviewed" keys
- Change status from "complete" to "incomplete" for all the cases where there's no ECSV file with spectral points or no YAML file with source measurements.

cdeil:
I've reviewed all `info.yaml` files and made further edits. Merging this now.

@pdeiml - Thanks!

PR #160, opened 2017-09-12
added info.yaml file for all relevant VERITAS papers

GernotMaier:
Relevant papers are all papers which are

- not of technical content only
- report a detection (exclude her all papers which reports upper limits only)
- gamma-ray papers only (e.g. no papers on cosmic-ray spectra are included)

The list of these VERITAS papers is complete.


cdeil:
@GernotMaier - Actually, empty `datasets:` keys aren't allowed by the schema, it has to be an "array", i.e. this would have worked: `datasets: []`. Instead of adding this empty array, I simply removed the empty `datasets:` keys for now in 4ea8698 , i.e. they should be added back once there's at least one dataset in gamma-cat for a given paper.

Thanks for adding this! Merging now.

PR #162, opened 2017-09-15
Issue157

pdeiml:
I reviewed everything twice, hence, you can directly merge it if you want.

Do you want me to merge all commits into one single commit?


cdeil:
I've fixed two small issues in 73c4ca0 and ef88fda .
For reference: I don't have time to review the diff for typos / data entry errors.
Merging now.

@pdeiml - Thanks!

PR #168, opened 2017-11-06
Add lightcurve to output 

pdeiml:
Modified collection.py:
Now, the lightcurve-ecsv-files are being copied to $GAMMA_CAT/docs/data as well.

The second commit updates the output files due to the upper changes.
@cdeil if you want to change the output files yourself, i can delete the "update output"-commit via git rebase.


pdeiml:
This is a PR which only adds the lightcurve-yaml-files to the output folder.

For the yaml-datasets and the input-index file. I will do another PR

PR #169, opened 2017-11-10
Basic source info on webpage

pdeiml:


pdeiml:
This PR became messy, hence, I opened
https://github.com/gammapy/gamma-cat/pull/172
which does in principle the same as this PR was intended to do.

PR #170, opened 2017-11-11
Sources to webpage

pdeiml:
This is a PR which adds a list of all sources to the documentation webpage.
Additionally, for each source the information about the existing sed-ecsv files is added.

I wanted to make two commit, one with the changes in webpage.py and one which adds the output of make.py webpage
to the documentation folder. But unfortunately, I made a rebase mistake, well, now it is one commit :-)

I will do follow up PRs in which all lightcurve-ecsv files will be added and some general information about the sources.

In principle, it is ready to merge and everything works as expected.
But I think, we need to discuss a little bit about the coding -> Inline comments


gammacat/utils.py
```diff
@@ -84,6 +84,17 @@ def fill_list(cls, data, key):
             return cls.fill_value['string']
 
 
+def convert_source_id(source):
```
cdeil: Zero-padding can be done with one line -> remove this helper function.
https://pyformat.info/#number
```
'{:04d}'.format(42)
```

gammacat/webpage.py
```diff
@@ -20,11 +23,60 @@ def __init__(self, *, out_path):
...
-        log.error('Come one. Implement me already, you lazy hog!!!')
+
+        self.make_source_list()
+        self.make_sources()
+
+    def make_source_list(self):
+
+        status_rst = str(gammacat_info.base_dir / 'documentation/data/src_list.rst')
+        #TODO: Shift the template things to a function in utils.py
+        templateLoader = jinja2.FileSystemLoader( searchpath = '/')
```
cdeil: On Slack I gave you the render helper function on Friday. Please add it to utils and then call it from webpage.py

gammacat/webpage.py
```diff
@@ -20,11 +23,60 @@ def __init__(self, *, out_path):
...
+        templateEnv = jinja2.Environment (loader = templateLoader )
+
+        #TODO: How to convert a PosixPath to normal string?
+        TEMPLATE_FILE = str(gammacat_info.base_dir / 'documentation/templates/src_list.txt')
+        template = templateEnv.get_template( TEMPLATE_FILE )
+
+        templateVars = {'data' : self.sources_data}
```
cdeil: Rename `templateVars` -> `ctx` (for "context", this is the name used commonly)

gammacat/webpage.py
```diff
@@ -20,11 +23,60 @@ def __init__(self, *, out_path):
...
+        #TODO: How to convert a PosixPath to normal string?
+        TEMPLATE_FILE = str(gammacat_info.base_dir / 'documentation/templates/source.txt')
+        template = templateEnv.get_template( TEMPLATE_FILE )
+
+        #TODO: Move this to another place? Maybe collection.py?
+        path = gammacat_info.base_dir / 'docs/data/gammacat-datasets.json'
+        gammacat_dataset = load_json(path)
+
+        for source in self.sources_data:
+            source_rst = sources_folder + '/source' + str(source['source_id']) + '.rst'
```
cdeil: Move the code to make one source page to a helper function.

pdeiml:
What do you mean with the auto-generated RST files?
I think, you mean the sourceXXXX.rst files in 
https://github.com/pdeiml/gamma-cat/tree/Sources_to_webpage/documentation/data/sources
These files are generated by make.py -> webpage but not with "make html".
I thought that the process from the master branch to the docu-webpage is in principle the same as executing "make html" on the local machine. But "make html" will not work if rat-files are missing, will it?

cdeil:
> But "make html" will not work if rst-files are missing, will it?

Correct.

Still: let's not put the generated RST files in the repo. One just has to run the two steps to make the webpage.

pdeiml:
Ok, I can add the "make html" command at the end of "make.py" webpage:
I have to look up the syntax but it would be something like sys.command(cd documentation && make html).

pdeiml:
This PR became a little bit messy, hence, I opened
https://github.com/gammapy/gamma-cat/pull/172
which does in principle the same as this PR was intended to do.

PR #173, opened 2017-11-15
Datasets to collection

pdeiml:
This is a PR which handles three things in the three commits: 

**1) Creation of the class DataSet**
The file gammacat/dataset.py is created which handles yaml datasets as class DataSet. It is written analogously to gammacat/lightcurve.py and gammacat/sed.py

**2) Add of dataset yaml files in 'input/data' to the output collection.** 
The new thing is that the starting point are the info.yaml files and not a scan through the input folder: 
Make.py scans trough the input folder and reads in the information given there. In particular, the yaml-datasets and data_entry['status'] as well as data_entry['reviewed']. 
Only datasets with data_entry['status'] == 'complete' are put to the output (see inline comment in collection.py).
This procedure can be expanded to the light curve and sed files which are currently read by a complete scan through the input folder. 
@cdeil l I think, we should do everything starting with the info.yaml file!

**3) Add light curves and datasets to output index file**
All light curve and dataset files in /docs/data/data are added to the function which produces the output index file.

**4) The output in /docs/ is updated**

@cdeil I am not sure whether the "docs" and "enhancement" label is correct for this PR :-)


gammacat/collection.py
```diff
@@ -211,6 +225,23 @@ def process_lightcurves(self):
...
             lightcurve.table.write(str(path), format='ascii.ecsv')
 
+    def process_datasets(self):
+        for filename in self.input_data.info_yaml_list:
+            folder = filename.parents[0]
+            info_data = load_yaml(filename)
+            status = info_data['data_entry']['status']
+            review = info_data['data_entry']['reviewed']
+            log.debug('Processing reference: {}'.format(info_data['reference_id']))
+            # TODO: This is if you want to make sure that all data are reviewed
```
pdeiml: Currently, only datasets, which are marked as 'complete' in the corresponding info.yaml file, are added/copied to the collection.
In the end, we should add a check for 'reviewed == yes' as well. But now, everything is not reviewed, hence the collection would be empty.

cdeil:
This is a large PR, I'll try to review this week, but it might have to wait for Monday.

The `process_datasets` method is too many for / if indentation, something should be split out.
I think we should expose all data we have, eventually showing which is reviewed or complete, but for now, just exposing and making all data we have visible is key.

cdeil:
For the issue / PR labeling: it doesn't matter at all for this repo. It only matters if you have a project with at minimum 5-10 contributors that don't talk to each other often. I just like the bright colors, that's why I sometimes put labels.

gammacat/dataset.py
```diff
@@ -0,0 +1,47 @@
...
+from .utils import load_yaml, write_yaml
+from gammapy.catalog.gammacat import GammaCatResource
+
+__all__ = [
+    'DataSet',
+]
+
+log = logging.getLogger(__name__)
+
+class DataSet:
```
cdeil: What is this class? Add a one-line docstring at least.

And please remove the one-line methods like source_id that just re-expose something from data. They don't help and just add more code to read and maintain.

pdeiml:
You are right, it is a quite large PR but a big part is only the output update . But this you do not have to really review, because it is basically only the copy of the input files.

I will wait until you reviewed everything.

gammacat/dataset.py
```diff
@@ -0,0 +1,47 @@
...
+        return self.data['reference_id'].replace('&', '%26')
+
+    def source_id(self):
+        return self.data['source_id']
+
+    def to_dict(self):
+        return self.data
+
+    def to_GammaCatResource(self):
+        try:
```
cdeil: this try except can be written equivalently in one line:
```
file_id = self.data.get('file_id', -1)
```
Why is a special case needed, what's -1?

gammacat/dataset.py
```diff
@@ -0,0 +1,47 @@
...
+    def folder(self):
+        return self.data['reference_id'].replace('&', '%26')
+
+    def source_id(self):
+        return self.data['source_id']
+
+    def to_dict(self):
+        return self.data
+
+    def to_GammaCatResource(self):
```
cdeil: Make this a property called `resource` to be equivalent with the LC and SED class?

pdeiml:
No, I did not add a input-index file only the datasets and lightcurves which are placed in /docs/data/data are added to the output index file.
But you are right. Index file for the input would be nice as well. I will do this in a future PR.

cdeil:
> Index file for the input would be nice as well. I will do this in a future PR.

Agreed! It's a bit change to the scripts, that will let us throw away ~ 50% of the code we have (e.g. in `input.py`), so it takes time to do together with at least some cleanup and review.

So just do the more minor code and docs suggestions I made inline and then this can go in very soon.

gammacat/dataset.py
pdeiml: Because in the light curve and see files file_id is set to -1 if there is only one file.

pdeiml:
I did two slight changes.
Unfortunately, I don't know how I should split the part in collection.py in a meaningful way.

Moreover, two general questions:
- I wrote the collection.py stuff in such a way that every dataset which is marked as "complete" is copied to docs/data. Is that ok or should we copy everything which means that files which are "incomplete" are copied as well. 
If we choose the second way then we should add a line to the file in docs/data that this file is incomplete (can be handled in the DataSet class).
- Should we rewrite the code which handles the light curve and sed files in analogy to this. I mean that the starting point is a index file created out of all info.yaml files and from that the copying to docs/data is handled.

I suggest to merge this PR now as it is and move the two questions above to an issue (just to remember it) and do another PR (linked to the issue). I think putting the upper two things in this PR, then it is a too big PR.

gammacat/collection.py
```diff
@@ -227,8 +270,26 @@ def make_index_file_for_output(self):
             resource = SED.read(self.config.path / filename).resource
             resource.location = filename
             resources.append(resource)
+        for filename in self.config.lc_files():
+            resource = LightCurve.read(self.config.path / filename).resource
+            resource.location = filename
+            resources.append(resource)
+        for filename in self.config.ds_files():
+            resource = DataSet(self.config.path / filename).to_GammaCatResource()
```
cdeil: https://travis-ci.org/gammapy/gamma-cat/builds/304770272#L1138

gammacat/collection.py
```diff
@@ -227,8 +270,26 @@ def make_index_file_for_output(self):
...
+            resource.location = filename
+            resources.append(resource)
+
+        for filename in self.config.lc_files():
+            resource = LightCurve.read(self.config.path / filename).resource
+            resource.location = filename
+            resources.append(resource)
+        for filename in self.config.ds_files():
+            resource = DataSet(self.config.path / filename).resource()
```
cdeil: I think here you need to remove the `()` at the end of the line because this is a property, no?

gammacat/dataset.py
```diff
@@ -0,0 +1,47 @@
...
+    'DataSet',
+]
+
+log = logging.getLogger(__name__)
+
+class DataSet:
+    """Process a dataset file."""
+    resource_type = 'ds'
+
+    def __init__(self, filename):
```
cdeil: Please use the coding pattern we use everywhere else, where `__init__` takes objects and a separate classmethod `read` does I/O.
https://github.com/gammapy/gamma-cat/blob/ebceb98cbc5beeaf305e5787e203300a2234fd87/gammacat/utils.py#L196

gammacat/dataset.py
```diff
@@ -0,0 +1,47 @@
...
+        log.debug('DataSet.__init__()')
+        self.filename = filename
+        self.data = load_yaml(filename)
+
+    def write(self, filename):
+        write_yaml(self.data, filename)
+
+    def reference_id(self):
```
cdeil: remove method, not useful.

gammacat/dataset.py
```diff
@@ -0,0 +1,47 @@
...
+        return self.data['reference_id']
+
+    def folder(self):
+        return self.data['reference_id'].replace('&', '%26')
+
+    def source_id(self):
```
cdeil: remove method, not useful.

gammacat/dataset.py
```diff
@@ -0,0 +1,47 @@
...
+        return self.data['source_id']
+
+    def to_dict(self):
```
cdeil: remove method, not useful.

gammacat/dataset.py
```diff
@@ -0,0 +1,47 @@
...
+        return self.data
+
+    def resource(self):
+        try:
```
cdeil: Replace try-except with this line:
```
file_id = self.data.get('file_id', GammaCatResource._NA_FILL['file_id'])
```

pdeiml: This does not work, don't know why. Hence, I leave it as it is.

pdeiml:
Ready to merge.

gammacat/dataset.py
cdeil: Works for me. I'm putting this now. If it doesn't for you, one cause might be that you have an old Gammapy. Also for the future: please post the error or say what / why this doesn't work, i.e. which output file isn't as expected.

cdeil:
This is good now! Just some minor code clean-up in 08291d30 .  Merging.

PR #175, opened 2017-11-22
Input index file

pdeiml:
This PR creates a index file with all resources in the input folder.
Currently, there isn't a distinction between the info.yaml keywords 'incomplete', 'complete' and 'reviewed' but all data are copied to the output collection (excepting the ones which are on status 'missing'). In the future, a distinction should be implemented. -> see the TODO in collection.py as well.

First commit adds a new variable 'in_path' to GlobalConfig and changes the existing 'path' variable to 'out_path' and adjustes all scripts where 'path' is used.

@cdeil I think at some point we should write collection.py completely new and split it into two main classes e.g. 'InputCollection' and 'OutputCollection' and a third one which provides functionality which is used in both Collection classes. 
E.g.  'make_index_file_output' and 'make_index_file_input' are basically equal and the code can be moved to this third class (or a class in util.py maybe???) and only call InputCollection.make_index() and OutputCollection.make_index() which are then both calling the code in the third class, hence, these two 'make_index' functions are only one line.

### DIFF
--- a/gammacat/collection.py
+++ b/gammacat/collection.py
@@ -6,7 +6,7 @@ import logging
 from collections import OrderedDict
 from pathlib import Path
 from astropy.utils import lazyproperty
-from gammapy.catalog.gammacat import GammaCatResourceIndex
+from gammapy.catalog.gammacat import GammaCatResourceIndex, GammaCatResource
 from .sed import SED
 from .lightcurve import LightCurve
 from .dataset import DataSet
@@ -22,6 +22,10 @@ __all__ = [
 
 log = logging.getLogger(__name__)
 
+# TODO: Rewrite this complete file:
+# E.g. three main classes 'CollectionConfig', InputCollection', 'OutputCollection'
+# because many things are used by input collection and output collection
+# hence we should find a way to not write code doubled
 
 class CollectionConfig:
     """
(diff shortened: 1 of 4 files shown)


pdeiml:
Ready to merge

gammacat/collection.py
```diff
@@ -256,27 +262,42 @@ def process_datasets(self):
...
+            if info_data['data_entry']['status'] == 'missing':
+                continue
+            # TODO: Decide which datasets are copied to output collection by the keywords in 
+            # 'status' and 'reviewed' in info.yaml
+            # e.g if info_data['data_entry']['status'] == 'complete':
+            for dataset in info_data['datasets']:
+                if dataset.endswith('yaml'):
+                    resource = DataSet.read(info_filename.parent / dataset).resource
+                elif dataset.endswith('ecsv'):
+                    try:
```
cdeil: This try except is no good, because `Lightcurve.read` might work on other ECSV files now, or in the future.
One way to do it would be to look for the "lc" or "sed" at the appropriate place in the filename string.

cdeil: Here's an example that shows that it currently doesn't work -- an SED is put as type: lightcurve in the index:
https://github.com/pdeiml/gamma-cat/blob/c91b69cf39d71811dfcc380082623791e74629f3/input/input-datasets.json#L13

pdeiml: Ok, but I don't know how to distinguish between a sed- and a light curve file.
I tried
elif dataset.endswith('lc*.ecsv')
....
elif dataset.endswith('sed*.ecsv')
...
but that does not work. I need this placeholder before the .ecsv because there can be multiple sed files in a folder.

cdeil: You should write a helper function, something like this:
```
def resource_type_from_filename(filename):
    return filename.split('.')[0].split('-')[-1]
```
To write such code, it's best to do it interactively in ipython
```
$ ipython
In [1]: filename = 'tev-000138-sed.ecsv'
```
OK, in this case it seems to be a one-liner, and if you only need it in one place in the code, you can just put the statement there. but if it gets more complex (e.g. because some filenames you pass there don't work with that expression and need to be filtered out) or if you use it multiple times, then it would be the time to split it out into a helper function.

pdeiml: Will not work.
For a file: tev-000034-sed-1.ecsv
filename.split('.')[0].split('-')[-1]
will lead to    1
while for
tev-000034-sed.ecsv
it will lead to   sed

So you would need something like
if ..endswith('.ecsv')
if filename.split('.')[0].split('-')[-1] == 'sed':
...
else:
if filename.split('.')[0].split('-')[-2] == 'sed':
... 
if filename.split('.')[0].split('-')[-2] == 'lc':
...

This will become extremely messy!

The problem is the "moving" position of the keywords 'lc' and 'sed' within the filename string!

pdeiml:
It is now a little bit messy, I know.
But it works and that is the most important thing, I think.

Ready for merging

gammacat/collection.py
```diff
@@ -258,16 +262,46 @@ def process_datasets(self):
...
+            # TODO: Decide which datasets are copied to output collection by the keywords in 
+            # 'status' and 'reviewed' in info.yaml
+            # e.g if info_data['data_entry']['status'] == 'complete':
+            for dataset in info_data['datasets']:
+                if dataset.endswith('yaml'):
+                    resource = DataSet.read(info_filename.parent / dataset).resource
+                    resource.location = str(info_filename.parent.relative_to(self.config.in_path) / dataset)
+                    resources.append(resource)
+                elif dataset.endswith('ecsv'):
+                    #TODO: Clean this messy stuff up!
```
cdeil: @pdeiml - Please give it your best shot to clean this up now before we merge. I already gave the suggestion that you should introduce a helper function. In addition, a lot of the lines you duplicate many times can be moved after the if / else.

pdeiml: Ok, I will give it a try this afternoon or tomorrow morning.

pdeiml:
I think the easiest and cleanest way to solve the inline-comment problem would be to divide 'datasets' in the info.yaml files into subsets of 'lc', 'sed' and 'yaml' files. Would look like that:

datasets:
   lightcurves: 
       tev-000049-lc-1.yaml
       tev-000049-lc-2.yaml
   spectra: 
       tev-000049-sed-1.yaml
       tev-000049-sed-2.yaml

Is a bit to do but - I think - doable within an hour

cdeil:
It seems a bit overkill if you can just get the dataset type from the filename with one line of Python.
If you want to change the input format, I would suggest to use something that's very close to our index files and extensible to add extra info per dataset, i.e. this:
```
datasets:
    - {type: 'lc', filename: tev-000049-lc-1.yaml}
    - {type: 'sed', filename: tev-000049-sed.yaml}
```

pdeiml:
Fixed it, ready to merge

cdeil:
Looks good now. Thanks!

PR #176, opened 2017-11-24
Fixup conf.py

pdeiml:
You can not access data on RTD, e.g. here
https://gamma-cat.readthedocs.io/data/sources/source_2.html
Clicking on the link to the data will lead to a non existing file.

I am not an expert on RTD and sphinx but I think the lines I changed are wrong.
Locally, it is enough to run 'make.py clean', then 'make.py webpage', followed by 'make html' in the documentation folder.
Hence, in the conf.py, it is not necessary to run 'make.py all' and the 'clean' and 'webpage' commands do not need the two hyphen.
@cdeil Please check whether it is working now (and if not, please fix it in some way)

### DIFF
--- a/documentation/conf.py
+++ b/documentation/conf.py
@@ -115,9 +115,12 @@ if not on_rtd:
 # http://docs.readthedocs.io/en/latest/builds.html#understanding-what-s-going-on
 if on_rtd:
     import subprocess
-    cmd = f'cd .. && {sys.executable} make.py all --clean --webpage'
-    print(f'Executing: {cmd}')
-    subprocess.call(cmd, shell=True)
+    cmd1 = f'cd .. && {sys.executable} make.py clean'
+    cmd2 = f'{sys.executable} make.py webpage'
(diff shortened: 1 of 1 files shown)


cdeil:
@pdeiml - This is not the right fix, it will give the same result.

The way RTD works is that the HTML build output folder `documentation/_build/html` and only that folder get copied to a webserver and served. Currently there is a broken link because at
https://github.com/gammapy/gamma-cat/blob/e570eaefc1cb05ea2e4d5beddcc1f0801873010a/documentation/templates/source_detail.txt#L15 you are linking to some other folder on the build machine. Since locally you use the same machine for building and serving the webpage it works, but not on RTD.

What we have to do is to put a copy of the output folder into `documentation/_build/html`, and link to the files there. @pdeiml - Do you want to try and change the scripts and line in the template I linked to above to do that? If you can't get it to work, assign the issue or pull request over to me and I'll try to find time to do it soon.

pdeiml:
I will try to fix it this week

PR #178, opened 2017-11-27
Fix path in rst-files

pdeiml:
I think, this should work for the webpage, shouldn't it?
gammacatinfo_basedir is platform indepent, I think.

Ready to merge (if it works)


cdeil:
This doesn't work. Please read my comment at https://github.com/gammapy/gamma-cat/pull/176#issuecomment-347101404 carefully. We need to put a copy of the data output folder in the Sphinx HTML output folder or it will not be available on the RTD webserver at all.

pdeiml:
Ok, then - I think - I don't understand the problem.
@cdeil Can you fix it?

pdeiml:
@cdeil Do you have time to fix this PR?

pdeiml:
What's the plan with this PR?

pdeiml:
In principle it should work. There is only a problem with the &-symbols in the paths which I cannot solve:

After executing "make html" in $GAMMA_CAT/documentation there are links to the datasets in the corresponding html-file of a specific source. Such a path in the html-file can contain a reference with "%26A" which is misinterpreted by my firefox because it wants to go to file which has a &-symbol instead of %26A.
All datasets without a &-symbol in the reference_id work fine, e.g. the ApJ data.

I don't know whether this is a general bug which needs to be fixed or only a problem of my local firefox browser.

@cdeil Can you have a look at it or merge it and look what happens on the webpage?!

cdeil:
I think we should eventually drive everything from `make.py`. Can you please add a `copy_data` method to the `WebpageMaker` and add it to the `WebpageMaker.run` call here?
https://github.com/gammapy/gamma-cat/blob/00c468150d20ef5774a8c0d5d1d9491bdec36c6d/gammacat/webpage.py#L41

Also, I'd prefer to have the folder be called "data" instead of "output", as I think that this is more intuitive for users (for them those files are the "input" they access).

Concerning the `&` handling, this should be done properly of course, so that it's always quoted when used in filenames and URLs. I thought we already do this?
https://gamma-cat.readthedocs.io/contribute/details.html#reference-identifiers
If no, it needs fixing. If you don't manage, please leave another comment and I'll look and fix this later this week.

pdeiml:
I tried to copy it via make.py with a command like this:
subprocess.call('mkdir documentation/_build/html', shell=True)
but this leads to an error. I think because of the underscore?

This &-symbol thing is very strange to me. jinja takes as an argument the location of a data-file with the correct %26-replacement and this path is correctly copied to the html-file by executing 'make html' locally. But as soon as I locally open the webpage, firefox interpreted the %26-symbol in the file path as a &-symbol and hence it does not find the file.
I don't know whether this is a bug of firefox or not.

cdeil:
> but this leads to an error.

In the future, please copy & paste the error, to increase the chance to get useful help.
Maybe you need to add the `-p` option to `mkdir` to make parent dirs also.
An underscore in a dir name isn't a problem.

Also just to make this dir, you don't really need to shell out, please do it from Python using this:
https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir
probably passing `parents=True` and `exist_ok=True`.

pdeiml:
Ok, I will try. Do we want to add the _build/output directory to the git repository?
If not, we have to make sure that RTD executes "make.py webpage" while building the website. I am not sure whether this is being implemented.

cdeil:
>  Do we want to add the _build/output directory to the git repository?

No

> we have to make sure that RTD executes "make.py webpage" while building the website

This is already implemented:
https://github.com/gammapy/gamma-cat/blob/00c468150d20ef5774a8c0d5d1d9491bdec36c6d/documentation/conf.py#L118

pdeiml:
Excepting this weird &-symbol problem described above, it is ready to merge/ it works.

The data folder in _build/html cannot be called "data" because "make html" creates a folder called "data" with html-files in it as well. Hence, you get a name conflict.
But I think it is fine like it is now because this is only something that is used by creating the webpage and not really used by contributors. And at some point we will rename "docs" to "output" as well and then everything is fine.

For me it would be very nice if you merge this, see whether the &-symbol problem is still there on RTD just to see wether the principle works.
Naming conventions can be changed later on!

cdeil:
@pdeiml - Yes, the URL encode issue was tricky.  See https://github.com/gammapy/gamma-cat/pull/178/commits/1da37e8339dcd229c6032c682899d3b70bbf083d#diff-2802bcc1a1e6a80826999cf02b2a2dd6R86

It's fixed for now and I think OK, although we could re-open the discussion if filenames should contain "&" in gamma-cat, or move the `url` property to a `GammaCatResource` property in Gammapy.

I accidentally commited output file updates also in 1da37e8339dcd229c6032c682899d3b70bbf083d, but they are small and look OK to me.

I'm merging this PR now, and if there's any issue, I can make a follow-up commit in master or a new PR.

cdeil:
Seems to work now, example:

* https://gamma-cat.readthedocs.io/data/sources/source_92.html
* https://gamma-cat.readthedocs.io/data/data/data/2008A%2526A...477..353A/gammacat_2008A%2526A...477..353A_000092_sed.ecsv

When I open the page locally, I see the ECSV file as text. When I click it online, the file downloads. Not sure what we want, probably offer both a way to display the data and download it. We should improve and discuss what is best one thing at a time and iterate towards a nice webpage.

First make sure all data is surfaced at all, then improve how it's displayed ...

PR #180, opened 2017-11-30
Add info and dummy file of HESS sources

pdeiml:
Add information and dummy files which are listed here:
https://www.mpi-hd.mpg.de/hfm/HESS/pages/home/sources/
but not in the input files.

Ready to merge


cdeil:
@pdeiml - Thanks!

There's one formatting error: https://travis-ci.org/gammapy/gamma-cat/builds/309572458#L1317 and there was a missing space in the HESS name. Fixed in 05631508 . Merging now.

PR #181, opened 2017-12-03
Add some VERITAS data

GernotMaier:
Pull requests contains additional VERITAS data.

- most incomplete data sets are due to missing light curve points (still working on that)
- completes a first pass through all VERITAS papers from 2008 to 2014 - there should be some information about every VERITAS paper for this period


GernotMaier:
Fixed the problem why checks failed - not sure how to rerun them.

cdeil:
@GernotMaier - Please leave a comment here when this is ready from your side, i.e. when I should have a look.

GernotMaier:
@cdeil  Fixed the problem with the error reported.

Please have a look now - this is now a first pass through all VERITAS paper. There are quite a few incomplete paper, which will take longer and deeper digging to get the values (mostly missing light curve points).

(I would run the make.py test program on my own, but I never got it running - should probably open a separate issue on that).

GernotMaier:
@cdeil Looks like all checks have passed - so this is ready for merging. Thanks for the patience

input/data/2017/2017A%26A...603A..31A/tev-000091-1.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+# VERITAS high (TeV flare  2009 May 1)
+spec:
+  erange: {min: 0.2, max: 6., unit: TeV}
+  mjd: {max: 54955., min: 54952.41}
+  model:
+    type: pl
+    parameters:
+      e_ref: {val: 1., unit: TeV}
+      norm: {val: 4.17, err: 0.24, scale: 1e-11, unit: cm-2 s-1}
```
cdeil: This should be in unit `cm-2 s-1 TeV-1`.
(I checked the paper: https://arxiv.org/pdf/1612.09472.pdf)

input/data/2017/2017A%26A...603A..31A/tev-000091-2.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+# VERITAS low state
+spec:
+  erange: {min: 0.2, max: 5., unit: TeV}
+  mjd: {max: 55004., min: 54907.}
+  model:
+    type: pl
+    parameters:
+      e_ref: {val: 1., unit: TeV}
+      norm: {val: 0.88, err: 0.06, scale: 1e-11, unit: cm-2 s-1}
```
cdeil: Same here: This should be in unit `cm-2 s-1 TeV-1`.
(I checked the paper: https://arxiv.org/pdf/1612.09472.pdf)

cdeil:
Clean-up commit: 7c2c650

There's one more problem that I'll file a separate issue.

Merging now.

PR #186, opened 2017-12-13
Hess papers

pdeiml:
Add two HESS papers and info about source 25.

### DIFF
--- a/input/data/2008/2008A%26A...481L.103A/info.yaml
+++ b/input/data/2008/2008A%26A...481L.103A/info.yaml
@@ -4,6 +4,7 @@ telescope: [hess]
 
 datasets:
   - tev-000008-sed.ecsv
+  - tev-000008.yaml
 
 data_entry:
   status: incomplete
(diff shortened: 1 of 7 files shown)


pdeiml:
I don't know where the fail of travisCI comes from, sorry.
Will have a look at that when I find time.

cdeil:
@pdeiml - You called this branch `HESS_data` and the one in #190 `hess_data`.
Don't do that, choosing names that just differ in case! I'm on Mac and it has case-insensitive filesystem, and was going crazy for the past 3 min, not understanding what's going on.

I'm merging this PR now via the Github interface, and then I'll review it after merging locally in the master branch.

cdeil:
I have reviewed and fixed something with this data entry
* for `input/data/2013/2013A%26A...554A..72H` in 1e8346e54a1ebc40562d96fefa3778152bf17104 (spectral index sign)
* for `input/data/2008/2008A%26A...481L.103A` in 79552568f3bacf79cb8a123d9354725b80e095f5 (morph type is point, this is an AGN, they have no measured extension even if the paper is a bit cryptic on that)

In both cases was the RA error incorrect, note that we use `Angle` to parse and `42s` means 42 arcsec, but for RA the error is `0h0m42s`, i.e. seconds, not arcseconds.

@pdeiml - Thanks for this contribution!

PR #187, opened 2017-12-14
Added yaml of V404 Cyg with number 168

micheledoro:
I was not sure what number should I use to add a new source and I used the next integer, after checking the source was not in the /input/sources. Don't know if it's correct. Will add data after

### DIFF
--- a/input/sources/tev-000168.yaml
+++ b/input/sources/tev-000168.yaml
@@ -1,0 +1,26 @@
+source_id: 168
+
+common_name: V404 Cygni
+gamma_names: []
+fermi_names: []
+other_names: []
+
+where: gal
+classes: [bin]
+
+discoverer: 
+seen_by: [magic,veritas]
+discovery_date: 
+
+tevcat_id: 
+tevcat2_id: 
+tevcat_name: 
+
+pos:
+  simbad_id:  
+  ra: 306.015958
+  dec: 33.867278	
+
+reference_ids:
+- 2017MNRAS.471.1688A
+
--- a/input/sources/tev-000168.yaml
+++ b/input/sources/tev-000168.yaml
@@ -8,19 +8,19 @@ other_names: []
 where: gal
 classes: [bin]
 
-discoverer: 
-seen_by: [magic,veritas]
-discovery_date: 
+discoverer: magic
+seen_by: [magic]
+discovery_date: 2099-12 
 
-tevcat_id: 
-tevcat2_id: 
-tevcat_name: 
+tevcat_id: 999
+tevcat2_id: 999-
+tevcat_name: 999-
 
 pos:
-  simbad_id:  
+  simbad_id: V* V404 Cyg
   ra: 306.015958
-  dec: 33.867278	
-
+  dec: 33.867278
+  
 reference_ids:
 - 2017MNRAS.471.1688A
 


GernotMaier:
Quick question: are we adding now as well non-detections / upper limits?

That would obviously mean adding a lot of sources.

cdeil:
@GernotMaier - I think we should allow the gamma-cat data collection to include non-detections / upper limits, and that we should take any VHE measurement / upper limit data  we get.

How about we add something like
```
detection_status: {'yes', 'no', 'maybe'}
```
[here](https://github.com/gammapy/gamma-cat/blob/00c468150d20ef5774a8c0d5d1d9491bdec36c6d/input/schemas/basic_source_info.schema.yaml#L37)?
Then it would be easy to filter on that. Maybe for the catalog file we only put the detections? Or keep also the non-detections? I don't have an opinion on that point yet, but for the data collection we want the upper limits.

@micheledoro - I think there's a formatting issue and your YAML file can't be parsed:
https://travis-ci.org/gammapy/gamma-cat/builds/316331018#L1025
I would suggest you try to trigger the error locally via `./make.py check` or directly via this command:
```
python -c 'from gammacat.utils import load_yaml; load_yaml("input/sources/tev-000168.yaml")'
```
And then fix the formatting issue in `input/sources/tev-000168.yaml`. Probably you have to replace tabs with spaces; maybe you also have to fix indentation after, using one of the other files as example. YAML like Python depends on proper indentation, and the main recommendation is to always use spaces and never tabs, so that indentation looks the same in any editor for anyone.

After that I'd be happy to merge this PR, and then the addition of `detection_status` or something similar to all the existing files could be done by @pdeiml or me in a follow-up pull request later this week.

micheledoro:
I will check the code, thanks.

Concerning Upper Limits, maybe we should move the discussion a different thread. However, I am pretty much in favour of adding them. We can even profit of gamma-cat to "publish" some unreleased observations (e.g. the 2h taken on target XX during alert YY which was not worth an individual publication). Of course, with this solution one would need to have a formal approval of the collaborations, and a tag expressing the  fact they are not "published in journal".

micheledoro:
Regarding this new version of the file, I found more than one place where I had to put dummy values:

discoverer: magic  <-- FALSE, IT IS AN NON-DISCOVERY
discovery_date: 2099-12 <-- FALSE, IT IS AN NON-DISCOVERY

tevcat_id: 999 <-- FALSE, IT IS AN NON-DISCOVERY
tevcat2_id: 999- <-- FALSE, IT IS AN NON-DISCOVERY
tevcat_name: 999- <-- FALSE, IT IS AN NON-DISCOVERY

So, before continuing with the data, I believe we should solve the terminology... In addition, if we continue in this direction, we should not ask for the tevcat_id...what we can do is add it in case the source is there, but I suggest we keep separated from tevcat...the fields of tevcats are in fact compulsory to fill

pdeiml:
I am not very familiar with measurement data, but can you tell me what data you would like to store in gamma-cat when you are talking about "upper limit"?

Is it a statement like "For energy X the flux is less than Y", hence, the two numbers to store would be X and Y?

GernotMaier:
I would say:
- MJD range, flux UL, non, noff, alpha (the latter three are important, as there are so many ways to calculate UL)
- for binaries, the MJD range is sometimes replaced by phases in orbital period (then we need to add also the assume orbital parameters to the yaml file)
- UL conf value
- if possible: add the method used for the UL calculation into the comment field

micheledoro:
@GernotMaier, when you say flux UL, do you mean integral from e_min to e_max or binned? About the comments, I believe it should be made explicit what kind of comments we want and maybe some values should be promoted to data such as confidence level, spectral index used to compute flux_UL from event_UL, if negative excess are rounded to zero or not, UL estimators (Rolke, F&C, etc). Without some of these, it is hard to reproduce a result accurately.

GernotMaier:
Maybe:
- Integral UL: only e_min is given
- binned: e_min and e_max is given

I agree with your other comments.

cdeil:
Note that for SED we already have a spec: http://gamma-astro-data-formats.readthedocs.io/en/latest/spectra/flux_points/index.html It supports upper limits for differential fluxed `dnde_ul` and integral fluxes `flux_ul`, and quite a few more things.
We don't have a scheme to declare yet more complex statistical measures or distributions, or a way to declare the method that was used to compute the upper limit. I think this could be added as a new keyword or just as a comment.

For light curves, we don't have the spec yet. But there was quite some discussion already, and there was consensus to re-use as much as possible the SED spec format. The remaining question before the LC spec can be written is how times should be stored exactly. If you're interested in this, please have a look at  https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/pull/61#issuecomment-349042750 and leave a comment.

Finally: note that we have to be agile and iterate. It's OK to add some data to gamma-cat, even if the spec isn't final yet, or to first work on the spec and add data later. Let the use cases from gamma-cat inform the spec and the other way around. Going through and changing the schema for the input folder in gamma-cat is annoying, but even for major changes where I changed the format for all spectral paramters in the YAML file a few month back, it only took a few hours.

GernotMaier:
@micheledoro @cdeil  
What was the conclusion on this discussion on how we include the papers with upper limits? Michele, did you add already an example?

My collection is:
- collect upper limits in sed or lc files (depending if they are differential or integral upper limits)
- add events counts (non, noff, alpha)
- tevcat_id, etc are not so important - so I am fine with Michele's placeholders. Better would be to simply leave it out.

Unclear for now how to add the method or anything complex to compute the upper limit. My suggestion is for now to include the method (and assumed spectral index, etc) in the comments field.

For the sources/...yaml files:
- do not include a 'discover' and 'discovery_date'
- replace the keyword 'seen_by' by 'observed_by'

I am not entirely sure about the detection_status: {'yes', 'no', 'maybe'}:
- it is need to count or find the detected sources
- but it will lead to discussions (as not every object which is claimed to be seen is actually detected). Who is deciding this?

micheledoro:
@GernotMaier @cdeil I agree with the comments from Gernot. As for the implementation, I prefer to leave it to you. One issue: if you leave now empty the discovery date, one cannot commit...

GernotMaier:
What was our conclusion on this topic? Do we want to follow the suggestions above?

(I cannot add any upper limits without adding 'fake' information on discovery until this is solved)

micheledoro:
Hi Gernot, nothing was decided to my knowledge.  One thing it is not clear to me is: would the addition of an 'option' in a field or a new field create a lot of confusion? I mean, all other files may need a modification? If so, the sooner a decision is taken the better. Second question: can we simply lift up the requirement that the field 'discoverer' is filled, so that some fields can be left empty?

Michele

cdeil:
@GernotMaier @micheledoro - Sorry for not being responsive here. I still don't have time to make a pull request adding something here, so for now I'll just leave a comment in case one of you has time.

I think we're all agreed that adding upper limit measurements in gamma-cat should be supported. That means we also have to support targets that aren't detected yet.

Another reason that I didn't comment here yet is because I'm not sure yet how to best structure the information. I think what we discuss above is fine, but it could also be structured in a different way of course, e.g. that would allow for more info, like who observed it when and whether they detected it:
```
observations:
- {by: hess, detected: yes, years: [2004, 2009]}
- 
```

For me, any addition that allows the non-detection use case is OK.

I think there's no way to avoid having something like `detection_status: yes / no / maybe`. This is not necessarily needed for the data collection, but when making a catalog of known TeV sources, that decision needs to be made. This has the potential to be controversial and adversarial, but we can just do one data entry / source at a time and take those decisions in the PRs that add the info. In the vast majority of cases we would simply follow the paper, if it says "discovery" in the title, that's what we put, if it doesn't then it's a non-detection. Of course there are a few bad and controversial claims, and for those we would put "maybe" and if someone complains re-discus.

cdeil:
> would the addition of an 'option' in a field or a new field create a lot of confusion?

I think it'll be fine, as long as we document what's what.

We currently have close to zero docs for users:
https://gamma-cat.readthedocs.io/

If anyone has time - pull requests to the webpage RST files that describe the data we have would be welcome any time:
https://github.com/gammapy/gamma-cat/tree/master/webpage

We do have the schemas that contain a description of the files and fields that we have:
https://github.com/gammapy/gamma-cat/tree/master/input/schemas
Simply linking to those would be a good start.

> I mean, all other files may need a modification?
> If so, the sooner a decision is taken the better.

Yes, we have to either update all input files, or add defaults to the processing of input -> output files in the Python scripts. Both things are some work, but it's not that bad. There's no way to avoid these "schema" changes in gamma-cat from time to time, because it's impossible to design a perfect, never changing schema that supports all data and use cases up front. You live and you learn (and put on some nice tunes while editing input files).

> Second question: can we simply lift up the requirement that the field 'discoverer' is filled, so that some fields can be left empty?

Yes, that is possible. I'm not sure if it is required at the moment. In any case, those things are done by editing the schema files like https://github.com/gammapy/gamma-cat/blob/11e570ae8a1d60c48f804f2908caca4197927234/input/schemas/dataset_source_info.schema.yaml#L273 or the Python scripts that process the input -> output files.

I'll try to come back and improve gamma-cat later in February, so @micheledoro @GernotMaier @pdeiml - either wait or go ahead if you have time.

GernotMaier:
OK - I agree with all your suggestions and I can wait a little bit longer for the implementations

micheledoro:
Hi @cdeil I also agree. I believe we can simply use the golden rule of the 5 sigma: >5 sigma 'detection', 3<sigma<5 'maybe', rest is 'no'. Further controversies will be probably based on scientific discussion, I hope, more than political one.

PR #190, opened 2017-12-15
HESS source data

pdeiml:
I added some information and dummy data files for HESS sources.


input/sources/tev-000065.yaml
```diff
@@ -1,10 +1,12 @@
+
 source_id: 65
 
 common_name: HESS J1418-609
 gamma_names: [HESS J1418-609]
 fermi_names: []
 other_names:
```
pdeiml: In the HESS source catalog HESS J1418-609 corresponds to Rabbit and G313.3+0.1 and not to Kookaburra.  -> See comment in source 66 as well

input/sources/tev-000066.yaml
```diff
@@ -4,7 +4,8 @@ common_name: HESS J1420-607
 gamma_names: [HESS J1420-607]
 fermi_names: []
 other_names:
-- PSR J1420-6048
```
pdeiml: In HESS source catalog HESS J1420-607 corresponds to Kookaburra -> See comment on source 65 as well

pdeiml:
ApSS 309 (2007) 277
This reference is mentioned for source 119 but I don't know whether we want to add it to the list in /input/sources/tev-000119.yaml or not.

input/sources/tev-000147.yaml
```diff
@@ -26,6 +26,8 @@ pos:
 reference_ids:
```
pdeiml: On the hess source catalog webpage there is also mentioned AICP 848 (2006) 570 as reference. Do we want to add this reference?

pdeiml:
Review required.

input/sources/tev-000065.yaml
cdeil: I think "Kookaburra" isn't really a well-defined thing. The best explanation I could find is in
https://www.aanda.org/articles/aa/full/2006/34/aa5511-06/aa5511-06.html
which says it's a complex region and both HESS J1418 and J1420 are in the Kookaburra region.
So I'm putting "Kookaburra" in `other_names` for both.

input/sources/tev-000066.yaml
cdeil: Thanks for spotting this. Yes, HESS J1420 should also have "Kookaburra" in `other_names`

input/sources/tev-000094.yaml
```diff
@@ -5,7 +5,7 @@ gamma_names: [HESS J1708-443]
 fermi_names: []
 other_names:
 - PSR B1706-44
-- G343.1-2.3
+- G343.1-1.3
```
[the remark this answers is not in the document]
cdeil: I think this is a typo in https://www.mpi-hd.mpg.de/hfm/HESS/pages/home/sources/
This is correct: http://www.physics.umanitoba.ca/snr/SNRcat/SNRrecord.php?id=G343.1m02.3
Changing back to "2.3" and sending an email to Konrad to fix the HESS source list typo.

PR #195, opened 2018-01-15
The big folder rename

cdeil:
This PR does the folder rename discussed in #171 :

`docs/data` -> `output`
`documentation` -> `webpage`

It also removes the old raw HTML page that was in `docs/data`.

I'm merging this now, and will adapt the scripts and documentation in a follow-up PR later today (the diff here is very large and hard to review, so I wanted to keep it separate)


cdeil:
I've disabled Github pages in https://github.com/gammapy/gamma-cat/settings

The old webpage at https://gammapy.github.io/gamma-cat is no more.

The new webpage is at https://gamma-cat.readthedocs.io/ and we'll continue to develop that one.

PR #199, opened 2018-01-15
Follow up of renaming

pdeiml:
I fixed the problem described at the end of 
https://github.com/gammapy/gamma-cat/issues/171

Additionally, I executed `make.py all` which lead to a update in the output folder.

### DIFF
--- a/gammacat/webpage.py
+++ b/gammacat/webpage.py
@@ -50,8 +50,8 @@ class WebpageMaker:
         """Copy output data folder to docs HTML output folder,
         so that the data files are available from the website
         """
-        src = gammacat_info.in_path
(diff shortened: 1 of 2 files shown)


cdeil:
Ah yes, I forgot to finish the update for the webpage and check if it works. That commit looks good to me.

The other one, where you update the output folder, looks just like you have different floating point formatting and that is not something that should be committed. Please remove that commit here, and update locally to Numpy 1.14, then you should get the same output that I committed earlier today.

pdeiml:
Ok, removed the update output commit.

RTM

cdeil:
Looks good. I didn't test locally. Merging now.

PR #200, opened 2018-01-15
Reorganisation of table of contents on webpage

pdeiml:
I reorganised the structure of the table of content which can be seen on the left side on RTD.
IMO it is a better to merge the data section and the user section and have only three main topics which describe gamma-cat in general (about), the user documentation and the contributor documentation, respectively.

Moreover, I added links to download the catalog files.


pdeiml:
Removed update output commit in https://github.com/gammapy/gamma-cat/pull/199

RTM

pdeiml:
Ok, you can merge it

cdeil:
Should the changes to `gammacat/webpage.py` be here? Or are they a duplicate from #199?

It's not clear to me if this is an improvement or not from looking at the diff.

Could you please post the updates HTML webpage somewhere for review? E.g. at your institute on a webserver, or push to a private repo you have and e.g. use https://rawgit.com to quickly have a static webpage.

Alternatively, sometimes just pasting screenshots into the PR discussion can be a nice way to show what changed.

pdeiml:
Ok, take a look now. 
As I said, it was a bit of a test. This webpage.py commit is the same as in 199 and I thought that as soon as you merge 199 this commit will disappear here. Unfortunately, it didn't.
But a short rebase fixed everything :-)

RTM

cdeil:
@pdeiml - please always share a link like in this case https://rawgit.com/pdeiml/html/master/html/index.html in the future for webpage preview after restructuring. And note that if you edit an existing comment, no new notification is sent, i.e. in this case I didn't see that you had a preview available like I asked, and thus didn't react here.

I'm very short on time this week, can't do a detailed review here. I think this is going in a good direction, merging your stuff now.

PR #201, opened 2018-01-17
Data for webpage

pdeiml:
Add the lightcurve ecsv files and the model yaml files to the list of source on the webpage.
Moreover, add galactic/ extragalactic to the source detail as a first additional information about the source. This needs discussion about what information shall be provided by the webpage and which not.

### DIFF
--- a/webpage/templates/source_detail.txt
+++ b/webpage/templates/source_detail.txt
@@ -21,4 +21,17 @@ SEDs
 Lightcurves
 -----------
 
-TODO
+{% for resource in resources.resources if resource.type == 'lightcurve' %}
+* `{{ resource.reference_id }} <../../output/{{ resource.url }}>`__
+{% else %}
+* None
+{% endfor %}
+
+Datasets
+--------
+
+{% for resource in resources.resources if resource.type == 'ds' %}
+* `{{ resource.reference_id }} <../../output/{{ resource.url }}>`__
+{% else %}
+* None
+{% endfor %}


cdeil:
I think this needs a rebase, no?

> Moreover, add galactic/ extragalactic to the source detail as a first additional information about the source. This needs discussion about what information shall be provided by the webpage and which not.

Do you already have a link to the "basic source info" YAML file?
I think exposing all info should have priority, and only then, maybe, re-expose parts of the info in a nice way in the HTML docs.
To start, let's either link to or completely include YAML and ECSV files as-is, to get all data visible and not spend a ton of time to discuss what to expose / script Jinja templates.

pdeiml:
Hi, indeed, it needed a review.
This part where I added the galactic or extragalactic info is more or less a test and starting point. Now, in the txt-template the starting point for information like that are there and in the future they can be changed easily.

Please review it and I think RTM

cdeil:
My suggestion was to link to the basic source info YAML file, so that people have access to that info. At the moment you are just exposing one key from that file, which is not very useful.

pdeiml:
Ah ok, then we had different suggestions.

I will do this yaml-source file link in a different PR because this needs a little bit more work, e.g. the source-yaml-file folder needs to be copied to _build/html.

So this PR is adding lightcurves and yaml datasets to the source list on the webpage.

RTM

cdeil:
Thanks!

PR #202, opened 2018-01-18
Publication list

pdeiml:
Add a section of all publications of which there is being data in gamma-cat.

The main commit of this PR is the second one "Add publication list to webpage". The other one is a copy of PR 201, hence, this PR here needs a rebase after merging PR 201.

My idea is that later on we will have a list of publications as links and for each publication all sees, datasets and lightcurves. Similar to the source list.

### DIFF
--- a/gammacat/webpage.py
+++ b/gammacat/webpage.py
@@ -9,6 +9,7 @@ from gammapy.catalog import GammaCatResourceIndex
 from .input import BasicSourceList
 from .info import gammacat_info
 from .utils import load_json, jinja_env
+from pathlib import Path
 
 __all__ = [
     'WebpageConfig',
@@ -44,7 +45,25 @@ class WebpageMaker:
         log.info('Make webpage ...')
         self.make_source_list_page()
         self.make_source_detail_pages()
+        self.make_publication_list()
         self.copy_data()
+
+    def make_publication_list(self):
+        publications = self.resources.unique_reference_ids
+        # The first entry of publications is empty, hence, delete first entry
+        del publications[0]
+
+        ctx = {
+            'publications' : sorted(publications)
+            }
+
+        template = jinja_env.get_template('publication_list.txt')
+        txt = template.render(ctx)
+
+        path = gammacat_info.webpage_path / 'use/publication_list.rst'
+
+        log.info(f'Writing {path}')
+        path.write_text(txt)
 
     def copy_data(self):
         """Copy output data folder to docs HTML output folder,
--- a/webpage/index.rst
+++ b/webpage/index.rst
@@ -31,6 +31,8 @@ This site is organised in the following sections:
    :caption: User Documentation
 
    use/index
+   use/source_list
+   use/publication_list
    stats/index
    changes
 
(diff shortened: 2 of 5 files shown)


gammacat/webpage.py
```diff
@@ -44,8 +45,26 @@ def run(self):
         log.info('Make webpage ...')
         self.make_source_list_page()
         self.make_source_detail_pages()
+        self.make_publication_list()
         self.copy_data()
 
+    def make_publication_list(self):
+        publications = []
+        for x in (Path(gammacat_info.out_path / 'data')).iterdir():
+            publications.append((x.parts[len(x.parts)-1]).replace('%26', '&'))
```
pdeiml: Is there a better way to decode the %26 to a &?

webpage/index.rst
```diff
@@ -31,6 +31,8 @@ This site is organised in the following sections:
    :caption: User Documentation
 
    use/index
+   use/source_list
+   use/publication_list
```
pdeiml: The subsections of user documentation have the same structure as in contributors documentation. One page per subsection and going the previous/next subsection by clicking on the arrows at the end of the page.
Think it is better to have one structure in the whole webpage.

gammacat/webpage.py
```diff
@@ -44,8 +45,26 @@ def run(self):
         log.info('Make webpage ...')
         self.make_source_list_page()
         self.make_source_detail_pages()
+        self.make_publication_list()
         self.copy_data()
 
+    def make_publication_list(self):
+        publications = []
+        for x in (Path(gammacat_info.out_path / 'data')).iterdir():
```
pdeiml: It is much more easier to get all references by scanning through the output/data folder instead of getting them by scanning through the output-index file. 
Firstly the entries in the index file are sorted by source_id and secondly the references can be doubled because there are more than one dataset for one reference.

cdeil: Please don't do that. The whole point of the index files it to avoid scrambling through directories from all over the place to see what's there. I think with the resource index helper class in Gammapy it's already possible to do anything you need here with one or two lines, i.e. select subsets of data as you like.

gammacat/webpage.py
cdeil: Yes, there is a helper function to quote and unquote the URL:
https://gamma-cat.readthedocs.io/contribute/details.html#reference-identifiers
Since we do that already when building the index files, here you should never have to do it.
But if so (e.g. to make a web URL), please use that helper function again and not raw string manipulation.

webpage/index.rst
cdeil: Please keep the separation in user and contributor documentation. The contributor documentation contains details about the formats, the Python scripts, Github, .... Users shouldn't have to worry about that.

pdeiml:
Ok, I rewrote the scanning through reference_ids with gammapy ResourceIndex.
Take a look and when it is fine, it needs a rebase against PR201 and then this here can be merged.

pdeiml:
Ok, it is ready to merge now.

@cdeil You are right, I will not do that multiple PR thing again. I just went into a rebasing problem that was quite tricky to solve. But - as mentioned yesterday - it was a test.

PR #208, opened 2018-01-25
Improve source details

pdeiml:
Add of 
- gamma-cat id to the source list on RTD,
- gamma-sky link to source details page
- link to the basic source info yaml-file
- three options for each dataset-/lightcurve-/sed-file: Show, Download and Gamma-Cat Input

If everything is fine, I would like to squash the commits


gammacat/webpage.py
```diff
@@ -114,6 +114,9 @@ def make_source_detail_page(self, source):
         # to the template render
         for resource in resources.resources:
             resource.url = urllib.parse.quote(resource.location)
+            resource.source_str = str('{:06d}'.format(resource.source_id))
```
pdeiml: This is a little bit hacky and there may be a better way to implement this.

webpage/templates/source_detail.txt
```diff
@@ -6,13 +6,21 @@ Source: {{ source.common_name }}
 Basic source info
 -----------------
 
-TODO
+{% if source.source_id < 10 %}
```
pdeiml: I tried to pass the source_id to jinja as a string like 000001 but jinja automatically converts it to a int.
These if-statements work but are not nice, I know.

webpage/templates/source_detail.txt
```diff
@@ -6,13 +6,21 @@ Source: {{ source.common_name }}
...
+{% if source.source_id > 100 %}
+`Information <https://github.com/gammapy/gamma-cat/blob/master/input/sources/tev-000{{ source.source_id }}.yaml>`__
+{% endif %}
 
 SEDs
 ----
 
 {% for resource in resources.resources if resource.type == 'sed' %}
-* `{{ resource.reference_id }} <../../output/{{ resource.url }}>`__
+* {{ resource.reference_id }} (`Show <https://github.com/gammapy/gamma-cat/tree/master/output/{{ resource.url }}>`__, `Download <../../output/{{ resource.url }}>`__, `Gamma-Cat Input <https://github.com/gammapy/gamma-cat/tree/master/input/data/{{ resource.year }}/{{ resource.parsedref }}/tev-{{ resource.source_str }}-sed.ecsv>`__)
```
pdeiml: @cdeil Do you have better names for "Download", "Gamma-Cat Input" and "Show"?

gammacat/webpage.py
cdeil: I think in many cases you can do formatting in the template.
Please read the Jinja template docs, especially this:
http://jinja.pocoo.org/docs/dev/templates/#format

There will still be some cases where things are pre-computed or pre-formatted before passing it to the template, and no hard line when to do which. It's up to you when you write it, with the goal to have it as simple as possible.

webpage/templates/source_detail.txt
cdeil: This is horrible. Please learn http://jinja.pocoo.org/docs/dev/templates and use that for formatting.

webpage/templates/source_detail.txt
cdeil: Suggest "Github input" and "Github output"

webpage/templates/source_detail.txt
```diff
@@ -22,7 +32,7 @@ Lightcurves
 -----------
 
 {% for resource in resources.resources if resource.type == 'lightcurve' %}
-* `{{ resource.reference_id }} <../../output/{{ resource.url }}>`__
+* {{ resource.reference_id }} (`Show <https://github.com/gammapy/gamma-cat/tree/master/output/{{ resource.url }}>`__, `Download <../../output/{{ resource.url }}>`__, `Gamma-Cat Input <https://github.com/gammapy/gamma-cat/tree/master/input/data/{{ resource.year }}/{{ resource.parsedref }}/tev-{{ resource.source_str }}-lc.ecsv>`__)
```
cdeil: Please break these long lines in the template, putting each url on a separate line, to make it more readable / maintainable.

Also, it would be better to pass a "github_base_url" to the template and render it, instead of to repeat the same base URL many times in the template.

gammacat/webpage.py
pdeiml: I played around with the formatting thing in python and jinja but I did not find a good solution, hence, I would like to leave it as it is.
It works and that's fine :-)

webpage/templates/source_detail.txt
pdeiml: See comment above

pdeiml:
I know that it is not very nice but I will not learn jinja in detail now. But with the time I will get more experience and then at some point I will be knowing how to format the template better and then I can fix it.

I would squash the commits and then it is ready to merge

cdeil:
@pdeiml - Then I'll just do the changes I suggested above early next week and then merge this in. Assigned the issue to myself just now.

Issue #198, opened 2018-01-15
Add year folder in output/data


cdeil:
I think the folder renames are done, except for one point:

In `input/data` we have sub-folders for each year, and in `output/data` we don't.

I think it would be easiest if the input and output folder had the same structure, i.e. we should introduce the year sub-folders in `output/data`.

From a quick look , what's needed is to remove most of the output filename building code in `CollectionConfig` and to just take the relative input location also as relative output file location.
https://github.com/gammapy/gamma-cat/blob/c028df0d0d06641afc0edef60fe9ed812078055a/gammacat/collection.py#L31

I don't have time for this now. @pdeiml - Either you make a PR for this (with just the scripting changes, not the file renames committed, so that it's easy to review), or I do it some day.
Let's here and generally work via Github issues, and just assign ourselves for issues when we start working on them, just in case we both start hacking on gamma-cat again.

pdeiml:
I agree with you that we should have similar structure in input and output. 
For me this change needs a bit of time and I don't know when I will have this time but I can do it.

cdeil:
I noticed that this is a major bug: we don't handle cases with multiple file_id when copying to the output folder, i.e. they are just discarded. Changing output folder / file structure to match input and then simply processing all resources via the index files will resolve this issue.

@pdeiml - Let me know if you have time for a PR to fix this week, or assign to me and I'll do it later this week.

pdeiml:
I started https://github.com/gammapy/gamma-cat/pull/215 which handles the ideas above. 
As soon as https://github.com/gammapy/gamma-cat/pull/215 is merged, this issue can be closed.

cdeil:
I think this issue is fully resolved. Closing See https://github.com/gammapy/gamma-cat/tree/master/output/data

PR #211, opened 2018-01-29
Move checks of info.yaml to input.py class

pdeiml:
I created two new classes in input.py, one representing the collection of the info.yaml files in /input/data and the second one representing one single info.yaml file.

This is much nicer than this dirty "check_info_yaml()" function which stood without any relation in checks.py.

The only thing which I am not sure about is the name of the classes.

RTM

### DIFF
--- a/gammacat/input.py
+++ b/gammacat/input.py
@@ -216,6 +216,47 @@ class BasicSourceList:
         log.info('Validating YAML files in `input/sources`')
         [_.validate() for _ in self.data]
 
+class InputInfo:
+    """All basic info about the data of a single publication.
+    """
+
+    schema = load_yaml(gammacat_info.base_dir / 'input/schemas/dataset_info.schema.yaml')
+
+    def __init__(self, data, path):
+        self.data = data
+        self.path = path
+
+    @classmethod
+    def read(cls, path):
+        path = Path(path)
+        data = load_yaml(path)
+        return cls(data=data, path=path)
+
+    def validate(self):
+        validate_schema(path=self.path, data=self.data, schema=self.schema)
+
+class InputInfoCollection:
+    """List of InputInfo objects.
+    """
+
+    def __init__(self, data):
+        self.data = data
+
+    @classmethod
+    def read(cls):
+        path = gammacat_info.base_dir / 'input/data'
+        paths = path.glob('*/*/info.yaml')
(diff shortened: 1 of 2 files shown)


cdeil:
@pdeiml - I'm very busy at the moment, Gammapy coding sprint next week and other project, and fixing #198 is much more important at the moment. I'll review this later.

gammacat/input.py
```diff
@@ -216,6 +216,47 @@ def validate(self):
         log.info('Validating YAML files in `input/sources`')
         [_.validate() for _ in self.data]
 
+class InputInfo:
```
pdeiml: This class handles one of the info files which are present in the input folder.
Basically, the only thing that it is being possible to do is to validate the file (but in the future there may be more functionality).

gammacat/input.py
```diff
@@ -216,6 +216,47 @@ def validate(self):
...
+    @classmethod
+    def read(cls, path):
+        path = Path(path)
+        data = load_yaml(path)
+        return cls(data=data, path=path)
+
+    def validate(self):
+        validate_schema(path=self.path, data=self.data, schema=self.schema)
+
+class InputInfoCollection:
```
pdeiml: This class is a collection of all info-files which are stored in the input folder. 
The only thing that it can be doing is to go through all files and validate each one using `InputInfo.validate()`. But - same as above - there may be more functionality in the future.

gammacat/input.py
```diff
@@ -306,6 +347,16 @@ class InputData:
     Expose it as Python objects that can be validated and used.
     """
 
+    def __init__(self, schemas=None, sources=None, datasets=None,
```
pdeiml: I shifted the __init__ function to the beginning of the class for better reading. Functionality is not changed (!).

gammacat/input.py
```diff
@@ -355,11 +398,14 @@ def read(cls):
         schemas = Schemas.read()
         sources = BasicSourceList.read()
         datasets = InputDatasetCollection.read()
+        ref_infos = InputInfoCollection.read()
```
pdeiml: `InputInfoCollection.read()` to have the collection of the info files as a variable of `InputDataCollection`. 
This is not being used but to have it similar to the other data files I added this line

PR #214, opened 2018-02-13
Update Webpage, no 1

pdeiml:
Restructuring of the table of content of RTD.
Additionally, some re-writing of the first section on RTD which gives an introduction to gamma-cat.

In a follow-up PR there will be a rewritten section for contributors.


webpage/index.rst
```diff
@@ -1,42 +1,31 @@
...
+   data/index
 
 .. toctree::
    :maxdepth: 1
-   :caption: User Documentation
 
    use/index
-   use/source_list
-   use/reference_list
    stats/index
```
pdeiml: What about this "stats" and "changes" parts. They are looking not very well but I don't know what the purpose is or was.

cdeil:
@pdeiml - Thanks! Very nice!

I've done some more docs edits in 05288d84

Merging this now.

If you plan to do further docs restructures, let's talk first to make sure we agree on the goal and don't go back and forth.

PR #215, opened 2018-02-13
New structure for output

pdeiml:
This PR handles the ideas from https://github.com/gammapy/gamma-cat/issues/198.

The new structure in the output folder is similar to the one in the input folder. The only difference is the "missing" info.yaml files which are not necessary in the output folder.

RTM


pdeiml:
I found a bug in my code. Please do not merge this here and wait until I write a new comment.

pdeiml:
I fixed it and squashed the three commits.
Now, it is ready to merge :-)

By the way: 
The sed and lightcurve files have "sed" and "lc" in the filenames in output. Do we want to add a "ds" to the name of the dataset files?

gammacat/info.py
```diff
@@ -58,15 +58,21 @@ class GammaCatStr:
...
-    def dataset_str(meta):
-        ss = GammaCatStr.reference_id_str(meta['reference_id'])
-        ss += '_'
-        ss += GammaCatStr.source_id_str(meta['source_id'])
-        return ss
+    def data_filename(meta):
+        return GammaCatStr.reference_folder(meta['reference_id']) + \
+            '/tev-' + GammaCatStr.source_id_str(meta['source_id'])
+                # if meta['file_id'] != -1:
+        #     return GammaCatStr.reference_folder(meta['reference_id']) + \
```
cdeil: Please always remove such old, uncommented code.

cdeil:
> Do we want to add a "ds" to the name of the dataset files?

I don't see any pros or cons, so would default to what we have.
If we were to do it, we would probably do it in input and output and it should be a separate discussion / proposal.

cdeil:
The travis-ci build is failing here:
https://travis-ci.org/gammapy/gamma-cat/builds/340908822#L2387

You can reproduce that error by running this locally:
```
python -m pytest -v --cov=gammacat
```

I think the fix is to update the reference value in the test here:
https://github.com/gammapy/gamma-cat/blob/176c191e20942526d659689375b74ff6cc0d502e/gammacat/tests/test_info.py#L9

pdeiml:
The diff is very easy, I think. You only have to look at the files in the gammcat folder which are collection.py and info.py. The rest is only running the output.
If you want to check whether it works you can check out this branch, go to output/data, delete everything there and then run "make.py collection". This will lead to exactly what is there right now because that is what I did. I deleted output/data and then I ran "make.py collection".

Concerning the TravisCI problem and your review, I will fix it later today or tomorrow morning

pdeiml:
I close this PR. #198 is handled in #216.

cdeil:
Thanks!  I'll review and merge #216 this afternoon.

PR #218, opened 2018-02-15
Add units in all ECSV files in input/data

cdeil:
As suggested by Gernot, we really should have units for every quantity, even in the input folder.
This PR adds it for all our ECSV files in input/data.


cdeil:
I've rebased this branch on master to get the better content in the output folder, that should now have all files. And only the LC files where the unit was missing changed a bit in the output folder: 28b1b63d
This shows that I didn't do any editing mistakes for all the SEDs in the input folder, but I put the units we had assumed as default before correctly.

Merging now.

PR #221, opened 2018-02-18
Quite a number of corrections to VERITAS data

GernotMaier:
This pull requests consist of a number of corrections for:

- wrong units
- wrong parameters
- missing files
- missing energy ranges
- wrong or mixup of values (or mix up of spectral points with errors)

I've cross-checked all existing spectral models and spectra for VERITAS - hope that the worst mistakes are corrected.


cdeil:
@GernotMaier - OK if I only give this a cursory review and then merge?

I think double-checking the numbers against the paper would take me a lot of time here.

GernotMaier:
Yes - that is fine. These are all 'obvious' mistakes.

(and I am trying to organise currently a file-by-file review inside VERITAS to cross check all values with the publications)

cdeil:
OK. We can / should also improve the scripts. With ~ 1 day or work we can have validation for a lot of data entry stuff, e.g. units.

input/data/2016/2016ApJ...821..129A/tev-000110-sed.ecsv
```diff
@@ -10,7 +10,7 @@
...
 # - telescope: veritas
 e_ref dnde dnde_err
-2.718 4.41e-10 1.76e-10
-5.119 7.18e-11 3.25e-11
-9.642 1.62e-11 7.11e-12
-18.16 4.15e-12 1.97e-12
+2.718 4.41e-14 1.76e-14
+5.119 7.18e-15 3.25e-15
+9.642 1.62e-15 7.11e-16
+18.16 4.15e-16 1.97e-16
```
cdeil: Just to confirm: I've checked this SED from VERITAS on G0.9 against the one from HESS
figure_1
https://github.com/gammapy/gamma-cat/blob/master/input/data/2005/2005A%2526A...432L..25A/tev-000110-sed.ecsv

Looks correct now, this was an error in 1e4 originating in m-2 vs cm-2 screw-up.

GernotMaier:
I've cross checked 2015ApJ...812...60B with the paper and the factor of 10 correction is needed.

This is the only VERITAS detection with such a problem. I've checked all VERITAS sources and for all the flux points and spectral points agree now.

PR #224, opened 2018-03-27
Fix typo: 2010A&A...511A..52A -> 2010A&A...511A..52H

pdeiml:
Fix of a bug in a input dataset file in commit 
https://github.com/gammapy/gamma-cat/commit/104296b07b678024897ded881704622240dbb6f5

This bug has lead to wrong information in the input- and output index files. 
Therefore a Update of the whole output folder is done in 
https://github.com/gammapy/gamma-cat/commit/7c1e0fef777d73d0c9c381f045830469520fcbfd,
additionally the input index file is fixed.

https://github.com/gammapy/gamma-cat/commit/7c1e0fef777d73d0c9c381f045830469520fcbfd
is a quite large diff because 'make.py all' changed a lot in the existing files.
I think this is due to a missing update of the output collection after changing scripts which affect the output collection.

I can drop the second commit but I do need the fix of this at least in the index files for the next PR!

RTM


cdeil:
> I can drop the second commit

Please do.

And in the future: please never add huge output folder updates to a PR that fixes something in the input folder.
Adding in related changes in the output folder is sometimes acceptable (but also not advised as it quickly leads to merge conflicts).
But adding a ton of changes to all the files that is completely unrelated is not.

Here, the diff you see for the output files is because you have an old version of Astropy (or the YAML package we use). Please update your Astropy and these differences should disappear for you locally.

cdeil:
I've applied an update to the output files in 939e3f3675d8043729888058e91d2eda265f7e58 .
Apparently I forgot to do that with the last gamma-cat PR, sorry.

But you see, now there's merge conflicts here in the output catalog files.
It's not possible to work with multiple PRs in parallel in gamma-cat if people commit unrelated changes to the output folder in PRs. That must be left up to the reviewer to apply them at the time of merging. Then we'll never have merge conflicts.

PR #225, opened 2018-03-27
Update reference detail on RTD

pdeiml:
This PR consists of three additions to the reference detail pages on RTD:

1) https://github.com/gammapy/gamma-cat/commit/43c7ab89a2090e9891457384712be7c69cc96cb4 adds the github input- and output folder to the list of basic information .

2) https://github.com/gammapy/gamma-cat/commit/f45cc6c60328f8f8d31b495638fc69a0f9539161 adds all GammaCat resources (dataset, sed, lc) stored in the output collection to the subsection 'Resources'.

3) https://github.com/gammapy/gamma-cat/commit/0815ce5385958e05c68efd0fd956a1c20b09096d adds the data entry status and the review status to the list of basic information.

The way the third change is implemented looks a little bit ugly and might need a bit clean up but for now it works and therefore it should be merged.

The diff is being very large because this PR is on top of https://github.com/gammapy/gamma-cat/pull/224. Hence, if https://github.com/gammapy/gamma-cat/pull/224 is merged the diff of this PR will be sweet and small :-)

RTM (after merge of https://github.com/gammapy/gamma-cat/pull/224)

### DIFF
--- a/gammacat/info.py
+++ b/gammacat/info.py
@@ -30,6 +30,10 @@ class GammaCatInfo:
         self.out_path = self.base_dir / 'output'
         self.webpage_path = self.base_dir / 'webpage'
 
+        # Input and output URL on github
+        self.input_url = 'https://github.com/gammapy/gamma-cat/blob/master/input'
+        self.output_url = 'https://github.com/gammapy/gamma-cat/blob/master/output'
+
         self.description = "An open data collection and source catalog for gamma-ray astronomy"
 
         self.datetime = str(Time.now())
--- a/gammacat/webpage.py
+++ b/gammacat/webpage.py
@@ -62,10 +62,21 @@ class WebpageMaker:
         references_data = []
         for reference_id in reference_ids:
             ads_url = f'https://ui.adsabs.harvard.edu/#abs/{reference_id}'
+            year = {reference_id[:4]}
+            in_folder = f'{gammacat_info.input_url}/data/{year}/{urllib.parse.quote(reference_id)}'
+            out_folder = f'{gammacat_info.output_url}/data/{year}/{urllib.parse.quote(reference_id)}'
+            data_status = (load_yaml(gammacat_info.base_dir / 'input/data' \
+            / f'{reference_id[:4]}/{urllib.parse.quote(reference_id)}' / 'info.yaml')) \
+            ['data_entry']['status']
+            review_status = (load_yaml(gammacat_info.base_dir / 'input/data' \
+            / f'{reference_id[:4]}/{urllib.parse.quote(reference_id)}' / 'info.yaml')) \
+            ['data_entry']['reviewed']
             references_data.append(dict(
                 reference_title_underline='-' * len(reference_id),
                 reference_id=reference_id,
                 ads_url=ads_url,
+                input_folder=in_folder,
+                output_folder=out_folder,
             ))
         return references_data
 
@@ -136,7 +147,6 @@ class WebpageMaker:
 
     def make_reference_detail_page(self, reference):
         ctx = {'reference': reference}
-
         template = jinja_env.get_template('reference_detail.txt')
         txt = template.render(ctx)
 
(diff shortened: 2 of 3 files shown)


pdeiml:
I might need to do a rebase after https://github.com/gammapy/gamma-cat/pull/224 is merged.
But if not and there are no merge conflicts, this PR is ready to merge :-)

gammacat/webpage.py
```diff
@@ -62,11 +62,26 @@ def make_references_data(self):
         references_data = []
         for reference_id in reference_ids:
             ads_url = f'https://ui.adsabs.harvard.edu/#abs/{reference_id}'
+            in_folder = f'https://github.com/gammapy/gamma-cat/blob/master/input/' \
```
cdeil: Can you please avoid the code duplication in these lines by putting the common things (e.g. the URL string, but also other lines look like copy & paste) into a temp local variable?

pdeiml:
I cleaned up the lines your inline comment referes to.
But I added a inline comment as well and please read it before merging.

gammacat/info.py
```diff
@@ -30,6 +30,10 @@ def __init__(self):
         self.out_path = self.base_dir / 'output'
         self.webpage_path = self.base_dir / 'webpage'
 
+        # Input and output URL on github
```
pdeiml: Line 33 - 36 might be not necessary -> see inline comment in gammacat/webpage.py

gammacat/webpage.py
```diff
@@ -62,11 +62,25 @@ def make_references_data(self):
         references_data = []
         for reference_id in reference_ids:
             ads_url = f'https://ui.adsabs.harvard.edu/#abs/{reference_id}'
+            year = {reference_id[:4]}
+            in_folder = f'{gammacat_info.input_url}/data/{year}/{urllib.parse.quote(reference_id)}'
+            out_folder = f'{gammacat_info.output_url}/data/{year}/{urllib.parse.quote(reference_id)}'
```
pdeiml: I think that 'gammacat_info.base_dir/input'  works instead of 'gammacat_info.input_url' as well (analogeously for output) . But I am not sure whether RTD handles this right.

Please test it on RTD and if it turns out that base_dir works then the changes in gammacat/info.py are not necessary (see my comment there).

cdeil:
To be honest, I'm not sure if I'll find time to do a detailed review / improvements to this code next week also ...

@pdeiml - I'm merging this now, trying it out locally.

cdeil:
I checked one example:
http://gamma-cat.readthedocs.io/data/references/2009ApJ...705.1624A.html

It looks good, but the following links are broken:

Edit on github: input/data/2009/2009ApJ…705.1624A
Catalog data on github: output/data/2009/2009ApJ…705.1624A

https://github.com/gammapy/gamma-cat/blob/master/input/data/%7B'2009'%7D/2009ApJ...705.1624A
https://github.com/gammapy/gamma-cat/blob/master/output/data/%7B'2009'%7D/2009ApJ...705.1624A

@pdeiml - You should see the same issue if you try this webpage locally. Can you please fix in a follow-up PR?

gammacat/webpage.py
```diff
@@ -62,11 +62,25 @@ def make_references_data(self):
         references_data = []
         for reference_id in reference_ids:
             ads_url = f'https://ui.adsabs.harvard.edu/#abs/{reference_id}'
+            year = {reference_id[:4]}
```
cdeil: The bug is here: you're puttiing a Python dict into year, but I think you just want the string, i.e. change to this:
```
year = reference_id[:4]
```

Also, if possible, we should avoid duplicated code that processes locations / urls and fiddles around with things like `reference_id[:4]` that now or in the future might not work for all reference ids.
I don't understand why you need to extract the year here, when we managed to avoid it for the source detail pages.
See `make_resource_info_for_template` as example.
The location should only be generated once when making the output index file, and after that you should use that existing field to make derived things like locations in the docs folder or URLs.

PR #226, opened 2018-03-29
Follow up of PR 225

pdeiml:
This is a follow up of https://github.com/gammapy/gamma-cat/pull/225 which fixes the github URLs on the reference details pages on RTD.

RTM


cdeil:
Small follow-up commit: fe2500563d10dcf5028bfa80b2804a6fc766140d
Output should be the same.
I tried to avoid code duplication and move a few things from the templates to `webpage.py` to keep the templates simple.

Issue #40, opened 2016-11-10
Add data for HESS J1718-374, G349.7+0.2, 2015A%26A...574A.100H


cdeil:
Folder already exists: https://github.com/gammapy/gamma-cat/tree/master/input/papers/2015/2015A%2526A...574A.100H
Paper: http://adsabs.harvard.edu/abs/2015A%26A...574A.100H
Spectral points: https://www.mpi-hd.mpg.de/hfm/HESS/pages/publications/auxiliary/auxinfo_SNRG349

cdeil:
I'll do this one today.

cdeil:
SED points added: https://github.com/gammapy/gamma-cat/blob/master/input/papers/2015/2015A%2526A...574A.100H/tev-000098.ecsv

I noticed that there's four upper limits at high energies in the paper, that are missing from:
https://www.mpi-hd.mpg.de/hfm/HESS/pages/publications/auxiliary/auxinfo_SNRG349
Emailing corresponding author now.

cdeil:
Update: added https://github.com/gammapy/gamma-cat/blob/master/input/papers/2015/2015A%2526A...574A.100H/tev-000098.yaml

cdeil:
Note to self: Diane Fernandez left science, Vincent Marandon and Cyril Trichard replied that they can't find the UL values any more.

So the 4 ULs should be read off the PDF Figure with a data picker.
Keeping this issue open as a reminder.

pdeiml:
What is the status about this issue?
The sed and yaml files are added and what about the upper limit thing?
By the way: Do we already add upper limits to Gamma-Cat?

pdeiml:
Added the upper limits in https://github.com/gammapy/gamma-cat/pull/228
Can be closed after merging https://github.com/gammapy/gamma-cat/pull/228

PR #228, opened 2018-04-05
Add HESSJ1718-374 sed data

pdeiml:
Add the missing upper limits of HESS J1718-374 sed data from 2015A&A...574A.100H.
In addition, the excess number is added to the yaml file.

https://github.com/gammapy/gamma-cat/issues/40 can be closed after merging this.

I reviewed all data in the folder of 2015A&A...574A.100H, hence, if @cdeil agrees, we can change the review status in the info.yaml file to "yes".

RTM

### DIFF
--- a/input/data/2015/2015A%26A...574A.100H/tev-000098.yaml
+++ b/input/data/2015/2015A%26A...574A.100H/tev-000098.yaml
@@ -5,6 +5,7 @@ telescope: hess
 data:
   livetime: 113h
   significance: 6.6
+  excess: 163
 
 pos:
   ra: {val: 17h17m57.8s, err: 0h0m2.0s, err_sys: 0h0m1.3s}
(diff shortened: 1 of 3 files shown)


input/data/2015/2015A%26A...574A.100H/tev-000098.yaml
```diff
@@ -5,8 +5,9 @@ telescope: hess
 data:
   livetime: 113h
   significance: 6.6
+  excess: 163
 
-pos:
+pos: #Check
```
cdeil: Remove this `#Check` comment; at least to me it's not clear to me what that means.

pdeiml: Ah, that was only for myself to mark that I reviewed the position against the paper.

pdeiml:
I added your remarks.

Unfortunately, I cannot merge by myself. Think I have no write access.
I will do a follow up PR with the related new output.

PR #230, opened 2018-04-06
Issue 40 update output

pdeiml:
Update of the output files produced by the new files in 
https://github.com/gammapy/gamma-cat/pull/228

RTM

@cdeil I think we should do a big output update because all sed files in the output folder are changed after 'make.py all'. I can add this here in a second commit :-)


pdeiml:
@cdeil What about the big update output?

cdeil:
@pdeiml - What do you mean by "big update output"?

I've re-run `make.py all` locally just now and committed new output files in 1795aa03
Seems like a small update to me.

PR #234, opened 2018-04-13
Clean up build of input index file

pdeiml:
This PR is on top of https://github.com/gammapy/gamma-cat/pull/211. Hence, you should firstly review and merge https://github.com/gammapy/gamma-cat/pull/211 before looking at the diff of this PR!

The plan is to have to classes `InputCollection` and `OutputCollection` each handling - as the names suggests - every data in the input and output folders, respectively.

This PR starts to write the `InputCollection` and it implements to first things:
1) A list of all `info.yaml` files (stored in input folder) is created via the class `InputInfoCollection` (see inline comment).
2) All of these info-files are validated (see inline comment). 
3) Secondly, the `input_index_file` is created after (!) the info-files are validated (see inline comment).
Step 2) and 3) are handled via a `run` function in `InputCollection` (see inline comment).

In the future these three steps should be the very first thing `make.py` does! This PR does not (!) change the procedure of `make.py`, it only introduces a ugly way to run the upper steps when `make.py collection --step input-index` is executed (see inline comment). There needs to be done a follow up PR which cleans up `make.py`.

The second small commit is only a small fix in the `_make_index_file_input` function so that it uses the `to_list()` function of `InputInfoCollection`. But the functionality is not changed.

### DIFF
--- a/gammacat/collection.py
+++ b/gammacat/collection.py
@@ -215,9 +215,7 @@ class InputCollection:
 
     def _make_index_file_for_input(self):
         resources = []
-        print(self.info_files)
-        print(type(self.info_files))
-        for info_filename in self.data.info_yaml_list:
+        for info_filename in self.info_files.to_list():
             info_data = load_yaml(info_filename)
             if info_data['data_entry']['status'] == 'missing':
                 continue
--- a/gammacat/collection.py
+++ b/gammacat/collection.py
@@ -12,7 +12,7 @@ from .sed import SED
 from .lightcurve import LightCurve
 from .dataset import DataSet
 from .info import gammacat_info, GammaCatStr
-from .input import InputData
+from .input import InputData, InputInfoCollection
 from .utils import write_json, load_json, log_list_difference, load_yaml
 
 __all__ = [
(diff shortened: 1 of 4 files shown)


gammacat/collection.py
```diff
@@ -199,6 +199,45 @@ def validate_list_of_files(self):
         expected_files = expected_files_extra + expected_files_sed
         log_list_difference(actual, expected_files)
 
+class InputCollection:
+
+    def __init__ (self, config):
+        self.config = config
+        self.data = InputData.read()
+        self.info_files = InputInfoCollection.read()
```
pdeiml: This line creates a list of all info files which are present in the input folder (step 1 in comment).

gammacat/collection.py
```diff
@@ -199,6 +199,45 @@ def validate_list_of_files(self):
...
+
+    def run(self):
+        self._validate_info_files()
```
pdeiml: Here, all info files which are present in the input folder are validated (step 2 in comment).

gammacat/collection.py
```diff
@@ -199,6 +199,45 @@ def validate_list_of_files(self):
...
+        self._make_index_file_for_input()
```
pdeiml: This line creates the input index file and stores it (step 3 in comment).

gammacat/collection.py
```diff
@@ -199,6 +199,45 @@ def validate_list_of_files(self):
...
+                    resource = DataSet.read(info_filename.parent / dataset).resource
+                elif dataset.endswith('ecsv'):
+                    if 'lc' in dataset:
+                        resource = LightCurve.read(info_filename.parent / dataset).resource
+                    elif 'sed' in dataset:
+                        resource = SED.read(info_filename.parent / dataset).resource
+                resource.location = str(info_filename.parent.relative_to(self.config.in_path) / dataset)
+                resources.append(resource)
+
+        self.index = GammaCatResourceIndex(resources).sort()
```
pdeiml: I am not sure whether this definition of the class instance variable `self.index` is pythonic or not. Currently, self.index is not used anywhere but I think in the future we will handle a lot of things via this index, hence, it is nice to have it as a class instance variable.

pdeiml:
I don't understand the error of travis ci. It is a http error?
`make.py` all does work locally, hence, there shouldn't be an error.

gammacat/collection.py
cdeil: If it's not used from elsewhere at the moment, better to keep it as a local variable only.

cdeil (commented):
@pdeiml -I merged #211 .

Should I merge this now, or are some edits neede first?

(I don't have time to review, I trust the change is a good one.)

pdeiml:
Honestly, I don't know whether you can merge this now after merging #211 :-D
I won't have time the next days but I will try to take a look to this again and then I give you feedback here.

PR #238, opened 2018-05-17
Update webpage

pdeiml:
Add simple information about https://github.com/gammapy/gamma-cat/blob/master/input/gammacat/gamma_cat_columns.yaml and https://github.com/gammapy/gamma-cat/blob/master/input/gammacat/gamma_cat_dataset.yaml to the user documentation of the webpage.

Moreover, I renamed the folder which contain that files. It is missleading to have a folder gammacat in the input folder because gammacat is either the python package or the catalog but this folder only contains some configuration files.

RTM


cdeil:
I would prefer to keep the shorter name. But OK, I don't care much, if you want the longer name.
@pdeiml - There's a merge conflict; can you please rebase.

PR #240, opened 2018-05-30
Issue 219 no2

pdeiml:
This is one way how to control which SED goes to the catalog.

If you look to input/gammacat/gamma_cat_sed.yaml and there at source_id = 1, you can see the idea:
- source_id and status is nothing new
- reference_id: all references where at least one sed-file of the source is stored
- catalog:
      - reference_id: The reference_id of the sed which goes to the catalog
      - file_id: If there are more than one sed file, the file with file_id goes to the catalog (if there is only one file, the default value is -1)

I only want to introduce this idea of a schema and if it is agreed then one needs to add the keywords to all sources.

### DIFF
--- a/gammacat/catalog.py
+++ b/gammacat/catalog.py
@@ -51,7 +51,10 @@ class DatasetConfigSource:
 
         For now, we always use the last one listed.
         """
-        return self.reference_ids[-1]
+        if (len(self.reference_ids) == 0):
+            return None
+        else:
+            return self.reference_ids[-1]
 
 
 class DatasetConfig:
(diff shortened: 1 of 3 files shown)


pdeiml:
Ok, if we stay with one single file, then I would suggest to add three additional infos:
sed_ref_id
lc_ref_id
ds_ref_id (data set)

What do you think?

cdeil:
I would leave `reference_id` as-is, and optionally if there is `reference_id_sed` or `reference_id_lc`, use that for SED or LC (and otherwise default to use the same for all).
I.e. minimal edits to get what we want.

But either way, your more verbose, but also more explicit choice for each part of the catalog data is also good.

